PR #710, opened 2026-07-08
feat(tasks): configurable default view (Today/All), defaults to All

h4yfans:
## What
Add a `defaultView` task setting (Today/All) so the Tasks page opens on the whole list by default instead of Today. Configurable in Settings → Tasks → Defaults; translated across all locales.

## Why
Opening on All matches how many people scan tasks — one less click to see the larger picture.


h4yfans:
Added tests covering the new \`defaultView\` setting: a Settings → Tasks select-change case (`tasks-section.test.tsx`) and a page fallback case for when a tab has no saved view state (`tasks.test.tsx`). Patch coverage for the changed lines is now covered.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

PR #745, opened 2026-07-15
fix(telemetry): report typed error codes and route out expected conditions

h4yfans:
## Production evidence (Loki, 2 days / 14 days on 2026.710.3)

Telemetry could not express what actually broke:

- **2x `NoteError`** (`Failed_to_update_note`) and **5x `SqliteError`** (`Failed_to_update_task`) — all un-triageable. `errorCode` was the *class name*, so the 7 typed `NoteErrorCode` values and better-sqlite3's `SQLITE_*` code were discarded. We could not tell `SQLITE_BUSY` (file locked) from a disk-full.
- **4x `Error` / `unhandled_rejection`** on 2 win32 installs with a **completely empty stack** — unactionable.
- **8x `Failed_to_list_Ollama_models`** — Ollama simply not running. **5x `Failed_to_connect_calendar_provider`** — the user abandoned the OAuth flow. Both are normal states reported as faults, drowning real signal.
- **Throttle masking**: `trackIpcError`'s throttle was keyed only by `error.name` and shared across *all* handlers, so a benign recurring `Error` masked a genuine different `Error` from another handler for 60s.
- **5x `SYNC_PAYMENT_REQUIRED` 402** in 2 days from one free user (4x `GET /sync/storage`), because `GET_STORAGE_BREAKDOWN` was not entitlement-gated client-side.

## The fix

**Typed error codes** — `toErrorCode` now prefers a typed code the error carries (`NoteError.code`, better-sqlite3's `error.code`, Node's `ECONNREFUSED`) over the class name. A note failure reports `NOTE_WRITE_FAILED`; a locked DB reports `SQLITE_BUSY`.

A code is only trusted when it looks like an enum token (`^[A-Za-z][A-Za-z0-9_.:-]{0,63}$`). Anything else — a path, an email, a URL, prose — is **rejected outright** and falls back to the class name. This is stricter than running it through `toSafeToken`, deliberately: `toSafeToken` would substitute characters and turn `/Users/kaan/secret.md` into `_Users_kaan_secret.md`, which passes `SAFE_TOKEN` but still leaks the path's structure. Every new code path is covered by explicit invariant tests (no `@`, `://`, `/`, `\`, ≤64 chars).

**Unhandled rejection stacks (main + renderer)** — a rejection reason can be any value, and a non-Error reason carries no stack. The likely cause of the empty-stack production lines: `buildErrorDetail` requires `error instanceof Error`, which **fails for a cross-realm error** (worker/preload), while `toErrorCode` still read `constructor.name` → `Error` — exactly the observed "`Error` with an empty stack". `normalizeRejectionReason` passes real Errors through, **adopts a cross-realm error's own frames**, and otherwise synthesizes a stack at the handler and names the reason's type (`Rejection_string`, `Rejection_Object`). The reason's message/value is never copied; `buildErrorDetail` still strips the stack header, so no message ships.

**Expected-condition routing** — the throw site marks the error, `trackMainError` skips marked errors. The decision is a pure, unit-tested predicate (`expected-conditions.ts`), following PR #733's `childProcessGoneErrorCode` precedent rather than being buried inline. Both cases still surface to the UI as an error envelope; only telemetry skips them.

**Exactly what is now suppressed** (deliberately narrow):
- Ollama model-list fetch failing with **`ECONNREFUSED` only** (walking the `cause` chain, incl. `AggregateError` for dual-stack localhost) = nothing listening on the port = Ollama is not running. A **DNS failure (`ENOTFOUND`), a connection reset (`ECONNRESET`), a bad HTTP status, or a JSON parse failure is still reported** — a real Ollama misconfiguration is not blanket-suppressed. There is a test asserting this.
- The **calendar OAuth timeout** error (the 10-minute abandoned-consent-screen path in `google/oauth.ts`). Only that one error object; every other calendar connect failure still reports.

**Throttle key** — now `action + errorCode`, same 60s window.

**Entitlement gate** — `GET_STORAGE_BREAKDOWN` copies the `GET_STATUS` precedent in the same file: `if (cached && !cached.isPaid) return null`.

`toSafeToken` / `toErrorCode` move into `packages/contracts/telemetry-api.ts`, replacing the byte-identical copy that main and renderer each carried, so both sides share one tested implementation.

## Compat / migration plan

No DB, no schema, no migration. **No contract shape change**: no telemetry event, field, or enum was added or altered — `TelemetryEventSchema` is untouched, so old and new clients emit the same shape and the server needs no change. The additions to `telemetry-api.ts` are pure helper functions.

- **Wire-compatible both ways**: `errorCode` was already a free-form safe token, so `NOTE_WRITE_FAILED` validates exactly like `NoteError` did. Old desktop builds keep sending class names; both parse.
- **No server change.** The 402 is correct server behaviour for a free user (Kaan confirmed) — only the pointless client call is removed.
- **`GET_STORAGE_BREAKDOWN` returns `null`** for known-free users — a value the handler *already* returned on the no-token path, so existing renderer callers handle it. Only a **known-unpaid** cached entitlement gates; an unknown/uncached entitlement still calls the server, so a paying user can never be locked out by stale local state (test covers this).
- **Loki/Grafana note (behaviour change, intended)**: queries pinned to `errorCode="NoteError"` / `"SqliteError"` will see those codes drop off as users upgrade, replaced by the typed codes. Dashboards filtering on them need updating. Suppressed Ollama/OAuth lines will disappear from error volume — expect a visible drop.

## Verification actually run

| Check | Result |
| --- | --- |
| `pnpm --filter @memry/desktop test:main` | **3209 passed**, 11 failed, 30 suites failed to load — all pre-existing, see below |
| 6 touched main suites (telemetry/ipc/oauth/sync-core) | **95 passed** |
| `pnpm --filter @memry/desktop test:renderer` | **5098 passed**, 6 skipped, 446 files — fully green |
| `packages/contracts` full suite | **1437 passed**, 44 files |
| `pnpm typecheck` | exit 0 |
| `pnpm lint` | exit 0 (3 pre-existing `no-base-to-string` warnings, none mine) |
| `pnpm check:architecture` / `check:contracts` | both passed |
| `pnpm ipc:generate && pnpm ipc:check` | IPC invoke map already up to date |
| `pnpm docs:impact --strict` + `pnpm docs:build` | both exit 0 |

**The two `test:main` failure groups are pre-existing and not from this change** — I verified rather than assumed:
- **11 failed tests**, all in `src/main/sync/manifest-check.test.ts` (a file I did not touch). I stashed my changes and re-ran it on a clean tree at `origin/main`: **11/11 fail identically**.
- **30 suites failed to load** with `Error: Electron failed to install correctly` — the known fresh-worktree native-deps issue, an environment problem affecting sync/agent suites regardless of this diff.

## TDD

Every test was written first and watched fail for the right reason. The throttle regression test is the highlight: it initially reported **0 of 2 expected calls**, because a bare `Error` thrown by an *unrelated earlier test* in the same file had already claimed the shared `Error` throttle key — the production masking bug reproducing itself inside the test suite.

## Not verified / risks

- **Not run live in Electron.** No real Loki line was observed end-to-end; the privacy and code-fidelity claims rest on unit tests, not a live event.
- The empty-stack root cause (cross-realm `instanceof` failure) is **inferred** from the code path matching the production signature, not proven against a win32 repro. The fix is strictly additive — it improves stackless reasons regardless of which mechanism produced them.
- `lastTrackedByCode` now holds more keys (action × code instead of code). Bounded by handler count × distinct failure modes, so small — but it is still an unbounded `Map` with no eviction, as before.

## Out of scope (follow-up)

`apps/landing` `BillingSection.tsx` fetches `canManageBilling` but never reads it → ungated "Manage payment method" button → 409s. Left untouched per the brief; noting it here so it is not lost.

## Coordination

`lib/errors.ts` untouched (owned by the sibling `fix-vault-write-error-cause` PR) — I only read `NoteError.code`. `childProcessGoneErrorCode` / `trackChildProcessGone` untouched (owned by `fix-utility-process-crash-attribution`). To keep those auto-mergeable, `diagnostics.ts` **imports** `toSafeToken` rather than deleting it, so the identifier stays in module scope and the sibling's `childProcessGoneErrorCode` body keeps compiling unchanged after a merge.

---

# Context for reviewers

## Where this came from

This PR is one of 8 opened from a **2-day production error-log triage** (2026-07-13 → 07-15, Grafana `/d/memry-logs`, `env=production`, Loki on the VPS).

The whole window held only **161 log lines** — 90 desktop errors, 27 server errors, 44 server warns — across **13 distinct installs**. Cloudflare Analytics Engine gives the denominator: **18 active installs** in those 2 days, **30** over 5 days.

Two things made this triage possible at all, and both are recent:
- **#732** (merged 2026-07-10) fixed `detectBuildChannel` falling back to `process.env.NODE_ENV`, which is undefined in packaged Electron. Before it, every packaged install reported `build_channel="development"` **and** defaulted telemetry OFF. Every desktop line in this window now reads `build_channel=production` on `2026.710.3` — so we are finally seeing real users.
- **#733** (merged 2026-07-10) stopped clean utility-worker exits from firing error events, so what remains is signal.

Low line counts here mean **low telemetry volume, not low impact**. The headline finding of the whole triage — a 58-day, 100%-broken attachment upload — was exposed by exactly **2 log lines**.

## How these PRs were produced

Each production error class was root-caused by a dedicated investigation against the real code before any patch was written, then implemented TDD-first in an isolated worktree, then put through an **adversarial review** whose explicit brief was to find what is wrong rather than to approve.

That review stage earned its keep: **6 of 8 PRs came back `needs-work`**, several with defects that would have merged silently. Its findings are reproduced verbatim below — they are not cosmetic.

**The one question that matters most on every one of these PRs:** *would the new tests actually fail on the base branch?* The 58-day attachment outage survived because both the server suite (R2/D1 fully mocked, self-consistent fixtures like `total_size: 10` with raw bytes) and the desktop suite (`fetch` mocked wholesale) passed green while the seam between them was never exercised. A test that mocks the thing under test proves nothing. Some of these PRs repeated that exact mistake and the reviewer caught it.

## This PR: production evidence

Our telemetry cannot express what actually broke. From the 2-day window:

1. **`errorCode` is the class name.** `toErrorCode` returns `error.name`, so every note failure collapses to `NoteError` (discarding the 7 typed `NoteErrorCode` values it already carries) and every DB failure to `SqliteError` (discarding the `SQLITE_*` code). We cannot tell `SQLITE_BUSY` from disk-full. Production: 2x NoteError, 5x SqliteError on `2026.710.3` over 14 days — all un-triageable.
2. **4x `unhandled_rejection` with a completely empty stack** on 2 win32 installs → unactionable.
3. **Expected conditions reported as errors, drowning real signal:** 8x "Failed_to_list_Ollama_models" (undici fetch failure because Ollama simply is not running) and 5x "Failed_to_connect_calendar_provider" (the user abandoned the OAuth flow).
4. **The throttle masks real errors.** `trackIpcError`'s Map is keyed only by `error.name` and shared across **all** handlers at 1 event/60s — so a benign recurring `Error` hides a genuine different `Error` from another handler.
5. **Free-user 402 noise:** 5x `SYNC_PAYMENT_REQUIRED` from one free user (4x `GET /sync/storage`, 1x upload initiate). `GET_STORAGE_BREAKDOWN` is not entitlement-gated client-side. **The 402 itself is correct server behaviour** — the fix is to stop the pointless client call, not to touch the server.

Note the throttle finding is not academic: it is the same class of defect as the 60s throttle potentially meaning the 66 `Utility:crashed` events in #741 **understate** the real crash rate.

## Related, deliberately not fixed here

`apps/landing` `BillingSection.tsx` fetches `canManageBilling` but never reads it → an ungated "Manage payment method" button → 409s. Different surface, own PR.

## Adversarial review findings (verbatim)

> Reproduced exactly as returned. Findings cite file:line — verify them yourself rather than trusting either the author or the reviewer.

VERDICT: needs-work

## TESTS ARE REAL?
MOSTLY YES, WITH ONE IMPORTANT EXCEPTION. The mock boundary is correct in the core suites: diagnostics.test.ts mocks only `./track` and telemetry-diagnostics.test.ts mocks only `./telemetry` — i.e. the downstream sink. toErrorCode, buildErrorDetail, normalizeRejectionReason and isExpectedConditionError all run for real. I traced each new assertion against the old implementation and convinced myself they fail on it: old toErrorCode returned `toSafeToken(error.name)` = 'NoteError'/'SqliteError', so the NOTE_WRITE_FAILED / SQLITE_BUSY assertions genuinely fail pre-fix; old renderer passed a raw string reason, giving errorCode 'StringError' and buildErrorDetail -> undefined (no `error` field), so the 'Rejection_string' + stack assertion fails pre-fix. The two validate.ts throttle tests are real and fail on old code (keys collapse to 'Error' / 'SqliteError' -> 1 call instead of 2). The privacy assertions (`not.toContain('secret')`, `not.toContain('@memrynote.com')`) are real end-to-end serialization checks, and I independently confirmed keepStackFrameLines strips the `Error: <message>` header so the adopted cross-realm stack is safe. EXCEPTION (blocking #2): the ai-inline 'still reports a real Ollama misconfiguration' test asserts on the marker predicate rather than on emission, so it is vacuous — it passes whether or not telemetry fires, and in fact telemetry does NOT fire because the previous test poisoned the live throttle key. The author explicitly presents this test as the guard against blanket suppression; it is not. Also note the sub-assertion 'and neither is a bad HTTP status' never exercises the `!res.ok` branch — fetch is mocked to REJECT, so the real `if (!res.ok) throw` at ai-inline-handlers.ts:108 is never reached. Separately, the cross-realm test at diagnostics.test.ts honestly documents `errorCode: 'Error'`, which contradicts the summary's claim of 'a code naming the reason's type' for that case.

### [1]
MASKING BUG REINTRODUCED — apps/desktop/src/main/ipc/validate.ts:20-27. `trackIpcError` writes `lastTrackedByCode.set(code, now)` BEFORE calling `trackMainError`, which then silently returns for expected-condition-marked errors (diagnostics.ts:49). So a SUPPRESSED error still claims — and refreshes — the throttle key. Combined with finding #3 (toErrorCode ignores `.cause`), I verified using the author's OWN test fixtures that ECONNREFUSED (suppressed) and ENOTFOUND (genuine) both yield errorCode `TypeError` and therefore the identical key `Failed to list Ollama models:TypeError`. Failure scenario: Ollama is not running (the common case, fires on every settings open/poll) -> each ECONNREFUSED claims the key and resets the 60s timer while emitting nothing; a genuine ENOTFOUND/ECONNRESET from the same handler within the window is then silently dropped. Because every poll refreshes the timestamp, a real misconfiguration can be suppressed indefinitely. This directly falsifies the PR's headline safety claim that 'a real Ollama misconfiguration is not blanket-suppressed'. Fix: call `isExpectedConditionError(error)` in trackIpcError and return before touching the Map, or have trackMainError report back whether it emitted.

### [2]
THE SAFETY TEST PROVES NOTHING — apps/desktop/src/main/ipc/ai-inline-handlers.test.ts, test 'still reports a real Ollama misconfiguration'. It asserts only `expect(isExpectedConditionError(dnsFailure)).toBe(false)` — it never asserts that telemetry was EMITTED for the DNS failure. It would pass even if trackMainError were never called, which is exactly what happens: this file does NOT mock `../telemetry/diagnostics` or `./validate`, so the real module-level throttle Map is live, and the PRECEDING test ('marks a connection-refused failure') already claimed key `Failed to list Ollama models:TypeError` at real time. The dnsFailure in this very test IS being throttled away right now and the assertion passes regardless. The author cites this test as evidence ('test asserts this', 'passed from the start — the correct guard against blanket suppression'); it passed from the start because it tests the marker function rather than the behaviour under test. This is precisely the self-consistent-mock failure the brief warns about. Must assert on the trackMainEvent/trackMainError sink being called with the DNS failure — and that assertion will FAIL until finding #1 is fixed.

### [3]
toErrorCode DOES NOT WALK `.cause`, SO THE ADVERTISED FIDELITY IS NOT DELIVERED — packages/contracts/src/telemetry-api.ts, `typedErrorCode` reads only the top-level `.code`. undici raises `TypeError: fetch failed` with the real code on `.cause`, so EVERY fetch/network failure still reports `TypeError` — no better than before. The commit message and PR summary both claim toErrorCode now prefers "Node's ECONNREFUSED"; that only holds for errors carrying `.code` at the top level (raw net/fs), not for the fetch path this PR actually touches. The author wrote a bounded cause-walker (`isConnectionRefusedError`, expected-conditions.ts:38-55) for exactly this nested shape, so the nesting was known — it just wasn't applied in toErrorCode. No test covers a nested cause: packages/contracts/src/telemetry-api.test.ts only exercises top-level codes. Either walk the cause chain in typedErrorCode (privacy is unaffected — TYPED_ERROR_CODE still gates the value) or drop the ECONNREFUSED claim from the commit message.

## NITS

- [1] PRIVACY INCONSISTENCY — packages/contracts/src/telemetry-api.ts, normalizeRejectionReason: `normalized.name = toSafeToken(name, normalized.name)` copies `.name` off an arbitrary non-Error reason using toSafeToken — the very function the author's own comment declares unsafe ('a character-substituted path (_Users_kaan_secret.md) still leaks its structure'). This is a NEW surface: old code hit `constructor.name` and returned 'Object'. Only reachable when the reason has stack frames (so it is error-shaped and `name` is conventionally a class name), hence low likelihood — but gate it with TYPED_ERROR_CODE like `.code` for consistency.

- [2] MISLEADING SYNTHESIZED STACK — for a non-Error reason, `new Error()` inside normalizeRejectionReason captures a CONSTANT stack pointing at normalizeRejectionReason -> trackMainUnhandledRejection -> index.ts. It says nothing about where the rejection originated but will read as a real stack in Grafana. The test only asserts `stringContaining('at ')`, which this trivially satisfies. The errorCode (Rejection_string etc.) is the actual win; consider dropping the synthetic stack rather than shipping a constant one.

- [3] TEST POLLUTION HAZARD — apps/desktop/src/main/ipc/sync-core-handlers.test.ts:426 uses `mockReturnValue({isPaid:false})`, which SURVIVES the `vi.clearAllMocks()` in beforeEach (clearAllMocks clears calls, not implementations). The reset to null sits at line 436 inside the test body AFTER the assertions, so any assertion failure leaks isPaid:false into the following pre-existing test 'fetches storage breakdown with a valid access token', causing a cascading failure. Use `mockReturnValueOnce` — the pattern this same file already uses at line 295.

- [4] SCOPE — the GET_STORAGE_BREAKDOWN entitlement gate (sync-core-handlers.ts:249-252) is product behaviour, not telemetry fidelity. It is defensible (it removes the SYNC_PAYMENT_REQUIRED noise the PR targets), the author flagged it, it copies a real in-file precedent (GET_STATUS at line 148), and it is backward-compatible — but it is the one hunk that does not trace cleanly to the brief.

- [5] STALE-ENTITLEMENT WINDOW — the gate trusts a persisted electron-store value. A user who upgrades on ANOTHER device has a stale `isPaid:false` cache and gets `null` storage breakdown until getBillingStatus refreshes it. The deep-link reconcile covers the same-device purchase path, and the identical precedent already exists in resolveEntitlementForSyncStart (paddle-billing.ts:116), so this is not new — just worth naming.


h4yfans:
## Update — review findings resolved, merge conflict resolved, ready for review

The adversarial review above (VERDICT: needs-work) has been actioned. All three BLOCKING findings and the two actionable nits are fixed TDD-first (each test was watched fail on base before the fix), and the branch was brought up to date with `main`.

**Merge conflict (was CONFLICTING → now MERGEABLE).** `main` gained #739, which added `NoteError.telemetryCode` (`NOTE_WRITE_FAILED:EBUSY` — code **plus** errno) and a `NoteError` special-case in `diagnostics.toErrorCode`. This PR moves `toErrorCode` into `packages/contracts`. Resolved by making contracts' `typedErrorCode` prefer a duck-typed `telemetryCode` over `.code`, so the errno fidelity survives **without** contracts importing desktop internals. #739's real-`NoteError` errno test is kept as the guard.

**Blocking**
1. **Throttle masking** (`validate.ts`): `trackIpcError` now skips expected conditions **before** touching the throttle Map, so a suppressed `ECONNREFUSED` can no longer claim/refresh the key and mask a genuine different failure. New `validate.test.ts` guard fails on base.
2. **Vacuous safety test** (`ai-inline-handlers.test.ts`): now asserts **at the telemetry sink** — `ECONNREFUSED` suppressed, a genuine `ENOTFOUND` reported — instead of only the marker predicate.
3. **`toErrorCode` ignored `.cause`** (`telemetry-api.ts`): `typedErrorCode` now walks `cause` / `AggregateError.errors` to a bounded depth, so an undici `fetch failed` `TypeError` surfaces the underlying `ECONNREFUSED`/`ENOTFOUND`. `TYPED_ERROR_CODE` gates every hop (no path/email/URL can be adopted at any depth). New cause-walk tests.

**Nits**
- `normalizeRejectionReason` now adopts a reason's `.name` only when it is an enum-ish token (reject, don't character-substitute a path).
- `sync-core-handlers.test.ts` gate test switched to `mockReturnValueOnce` so a failure cannot leak `isPaid:false` into the next test.

Left as-is (non-blocking, deliberate): the synthesized rejection stack (#2), the `GET_STORAGE_BREAKDOWN` gate scope (#4), and the stale-entitlement window (#5).

**Verification:** contracts 50/50 · desktop main (touched suites) 69/69 · renderer telemetry 5/5 · `pnpm typecheck` 16/16 · `pnpm lint` 0 errors · `pnpm ipc:check` up to date · `docs:impact --strict` covered · `docs:build` ✓. A second adversarial review pass over the fixes returned merge-ready with no blockers.

PR #757, opened 2026-07-16
feat(calendar): drag tasks onto the calendar to schedule them

h4yfans:
## Drag tasks onto the calendar to schedule them

Schedule and reschedule tasks by direct manipulation, from three sources:

1. **Unscheduled tab** (new) in the Day Panel — lists tasks with no due date; drag a row onto any calendar cell to schedule it.
2. **A Tasks pane beside the calendar** in split view — drag a task row straight onto the grid.
3. **Task chips already on the calendar** — drag to a new day (month) or a new day/time (week/day grid).

### Drop semantics

| Drop target | Result |
| --- | --- |
| Month day cell | Sets due date; keeps existing time |
| Week/day all-day row | Due that day, time **cleared** |
| Week/day time slot | Due at dropped time, snapped to 15 min |

The all-day row reveals itself during a task drag even on weeks with no all-day items, so there's always somewhere to drop to clear a time. Multi-selected tasks drag together and schedule in one drop.

### Architecture

Drag responsibility splits by **granularity**, not by source:

- **dnd-kit** owns date-granular drags (rail, cross-pane, all-day, month) landing on `type: 'date'` droppables. This completes the dormant `'date'` drop-target scaffolding that already existed in `drag-context.tsx`.
- The existing bespoke **`useEventDrag`** owns time-granular moves on the week/day time grid, and is now admitted for timed task chips (event move/resize behavior is byte-unchanged).

Both converge on `handleDateDrop` → `onUpdateTask` → the tasks domain, so sync bumps only the changed field clocks.

### Backward compatibility

Single additive change to the contract: an optional `unscheduled` boolean on the task-list schema. **No schema change, no migration, no sync-protocol change.** `dueDate` + nullable `dueTime` is the existing model — no new fields.

### Verification

- `typecheck`, `lint` (0 errors), `ipc:check`, `check:architecture`, `check:contracts`, `git diff --check` — all green.
- Renderer test suite: **5145 passed, 0 failed**. New unit + integration tests throughout; several assert the real dnd-kit registration and cross-seam behavior (not cosmetic DOM attributes).
- Main-process change (unscheduled query) `tasks.test.ts`: 24/24.
- Docs gate green; VitePress docs build clean. Docs added to `calendar.md` and `day-panel.md`.
- **Live in-app QA not yet run** (Electron install issue in the dev worktree) — the one reason this is a draft. Everything below the app runtime is verified.

### Review rigor

Built task-by-task with an adversarial review after each and a final whole-branch review. Findings that were caught and fixed rather than shipped:

- Six tests across the branch that couldn't fail against a broken implementation — rewritten to assert real behavior, each with a verified mutation kill.
- The riskiest task (week/day drop wiring) was sent back when a reviewer sabotaged the view and all 255 tests still passed; closed with wiring tests that fail under that exact mutation.
- Final whole-branch review found a real data bug: dropping a **timed** task onto an all-day cell didn't clear its time (`use-task-queries.ts` mapped `dueTime: null` to `undefined`). Fixed across all three update paths — this also fixes a pre-existing bug in the task-drawer time picker's "clear" action.

### Known limitation / follow-up

**Undo for dnd-kit calendar drops is not wired.** Moving a timed task chip on the grid is undoable with Cmd/Ctrl+Z (via the calendar's existing undo path). But drops from the Unscheduled tab / cross-pane / all-day / month record an undo action into `useDragHandlers`' internal stack — which is **dead app-wide today** (`App.tsx` never consumes the hook's `undo`), so tasks-page section-drop reschedules are equally un-undoable. Wiring it correctly is non-trivial (the stack-based `undo` would capture a stale stack in a toast closure) and touches shared, pre-existing code, so it's called out here for a deliberate decision rather than done hastily in this PR. Docs describe the actual behavior.

### Spec & plan

- Design: `docs/superpowers/specs/2026-07-15-calendar-task-drag-schedule-design.md`
- Plan: `docs/superpowers/plans/2026-07-15-calendar-task-drag-schedule.md`


apps/desktop/src/renderer/src/pages/calendar.tsx
```diff
@@ -687,7 +689,45 @@ export function CalendarPage({ className: _className }: CalendarPageProps): Reac
     await queryClient.invalidateQueries({ queryKey: ['calendar', 'range'] })
   }
 
+  const dueDateTimeFromDate = (date: Date): { dueDate: string; dueTime: string } => ({
```
h4yfans: Fixed in 43aa2c6ef — hoisted `dueDateTimeFromDate` to module scope beside the other calendar helpers. It closes over no component state (only `formatDateKey` + its `date` arg), so this is a safe no-behavior-change move. Typecheck + calendar tests green.

PR #765, opened 2026-07-16
chore(desktop): upgrade Electron 39 → 43.1.1 + adopt safe 40-43 wins

h4yfans:
## Summary

Upgrades Electron **39.8.6 → 43.1.1** (Chromium 142→150, Node 22.20→24.18, V8 14.2→**15.0**) and folds in the safe, inert modernizations that Electron 40–43 unlock. Larger, higher-risk adoptions are deliberately **sequenced into follow-up PRs** (see roadmap below) rather than crammed onto a risky runtime bump — this is a live beta with real user data, so the bump lands green and self-contained first.

Fresh work off `main`; the closed #719 was used only as a known-good reference.

## Dependency changes

| pkg | from | to | why |
|---|---|---|---|
| `electron` | ^39.8.6 | **^43.1.1** | target; carries the `ValidateIntegrityOrDie` asar-swap auto-update crash fix |
| `@electron/rebuild` | ^4.0.3 | ^4.2.0 | ABI 148 support |
| `better-sqlite3` | ^12.5.0 | ^12.11.1 | desktop + sync-server + app-core; 12.6.x won't compile against V8 15 |
| `electron-updater` | ^6.6.2 | ^6.7.3 | align range with installed; E43 auto-update behavior |
| `node-abi` (override) | 4.26.0 | **4.33.0** | pinned via `overrides` + `minimumReleaseAgeExclude` — the transitive `node-abi` was capped at 4.26.0, which predates Electron 43 (`Could not detect abi for version 43.1.1`) |

## Compatibility fixes

- **White-window fix** (`packages/i18n/src/shared/direction.ts`): V8 15 removed the legacy `Intl.Locale#textInfo` getter; reading it threw at renderer boot → blank window. Now feature-detects `getTextInfo()` with a `.textInfo` fallback. *Only E2E on bundled Electron catches this — vitest's V8 still has the old shape.*
- **memry-file CORS** (`src/main/index.ts`): Electron 43 blocks cross-origin `fetch()`/XHR against custom schemes registered with `supportFetchAPI` unless `corsEnabled` is also set. Added it as a forward-safety net (all current access is via no-cors `src` loads; keeps `fetch(memry-file)` byte reads working).

## Safe 40–43 adoptions (folded in)

- **Notification `failed` handler** (`reminders.ts`): E42 routes macOS notifications through `UNUserNotificationCenter`; unsigned builds emit `failed` instead of showing. Logged for diagnosability (reminders already back-stop as inbox items). Signed production builds are unaffected.
- **Notification `id`/`groupId`** (`reminders.ts`): stable per-reminder id + shared group so same-tick banners collapse in Notification Center (E42; inert on <42 and Linux).
- **`memory-eviction` GPU guard** (`index.ts`): E40 adds a `memory-eviction` child-exit reason (OS memory-pressure kill); it's now treated as benign so it no longer misfires the GPU-crash guard and needlessly disables hardware acceleration next launch.
- **`serviceName` on the CRDT preflight fork** (`crdt-preflight.ts`): a native abort there now surfaces as `CrdtPreflight` in `ps`/telemetry instead of an anonymous `Node Utility Process`.
- Free startup-perf win comes with the bump (Node snapshot boot + preload V8 bytecode cache) — no code.

## Test harness

E43's faster startup exposed a latent race in `crdt-preflight-crash.e2e.ts`: it force-crashes the preflight child and scrapes `mainLogs` for the failure report, but that log is emitted during early startup — **before** the window paints and before Playwright's stdout capture attaches, so faster boot means it's never in `mainLogs`. Added `waitForMainLog`, which also polls electron-log's on-disk file (the reliable source for pre-window logs). Verified the app itself is correct under E43 (preflight runs → child aborts → failure detected → app degrades to in-memory, window opens).

## Verification

- `lint` · `typecheck` · `ipc:check` — green
- desktop **vitest 10949 pass / 0 fail** (`--no-file-parallelism`; the parallel-worker SIGSEGV is a pre-existing environmental flake, not a regression — desktop tests are informational on `main`)
- sync-server + 4 workspace packages — green
- **E2E on bundled Electron 43** (`--workers=1`): boot-critical subset **36 passed** — i18n/RTL boot (the `getTextInfo` path), notes CRUD, memry-file attachments/import, crdt-preflight crash+degrade. CI runs the full matrix on macOS/Windows/Linux.
- `docs:impact --strict` covered · `docs:build` green

## Follow-up roadmap (ambitious 40–43 wins, sequenced — not this PR)

Deliberately separate PRs after this bump is green, because bundling irreversible data migrations / integrity fuses / sandbox flips onto a runtime bump risks a fleet break with no client-side hotfix:

- **A — keytar → `safeStorage` (sync API):** dual-read migration across the 5 keychain call sites; gate on `isEncryptionAvailable()`, refuse to drop keytar on the Linux plaintext backend, decrypt-round-trip verify before deleting keytar, keep keytar read-only fallback ≥1 release. *Irreversible master-key migration — its own PR.*
- **B — `safeStorage` sync → async** (`encryptStringAsync`): behavioral-only, same on-disk ciphertext; after A is stable in the field.
- **C — security hardening:** `@electron/fuses` + ASAR integrity (reconcile with electron-updater's asar swap), deny-by-default permission handler (allowlist `media` + iframe needs), `will-frame-navigate` guard, renderer `sandbox: true` — each behind a full 3-OS E2E pass.
- **D — notification polish:** `Notification.remove` on dismiss/snooze (uses the id/groupId landed here), dock `setBadgeCount` wiring.
- **Evaluated & rejected:** `globalShortcut.setSuspended`, and speculative APIs (`utilityProcess disclaim`, `configureWebAuthn`, `webContents.clone`, `setBackgroundBlur`, MSIX) — none fit the codebase.


h4yfans:
Follow-up roadmap tracked in https://github.com/memrynote/memry/issues/770 (milestone **Electron upgrade**): #766 (keytar→safeStorage), #767 (async), #768 (security hardening), #769 (notification polish).

apps/desktop/tests/e2e/utils/electron-lifecycle.ts
```diff
@@ -123,6 +123,47 @@ export async function destroyElectronApp(app: ElectronApplication, dirs: string[
...
+  timeoutMs = 15_000
+): Promise<boolean> {
+  const candidates: string[] = []
+  try {
+    const dirs = await launched.app.evaluate(({ app }) => ({
+      logs: app.getPath('logs'),
+      userData: app.getPath('userData')
+    }))
+    // electron-log v5 writes to getPath('logs'); older convention is userData/logs.
+    candidates.push(path.join(dirs.logs, 'main.log'), path.join(dirs.userData, 'logs', 'main.log'))
```
h4yfans: **Stale-log false-green on macOS** — _(fix applied in the working tree)_

`waitForMainLog` polls `app.getPath('logs')/main.log`, but on macOS that resolves to the shared `~/Library/Logs/<appName>/main.log`, which lives **outside** the per-test temp `--user-data-dir`. This test force-crashes the preflight on every run and appends `"CRDT store failed preflight"` to that shared file, so a line left by *any prior run* satisfies the assertion even when the current launch never logged it — silently defeating the "guard against a vacuous pass" (lines 40-45 of the test) that this call exists to provide. A future regression where the app no longer degrades/logs would still go green.

Linux/Windows resolve `logs` under the temp userData, so they're already isolated per-run; macOS — the dev platform and part of the CI matrix — is the exposed one. (electron-log v5 docs: `libraryDefaultDir` = `~/Library/Logs/{appName}` on macOS vs `userData/logs` on Linux/Windows.)

**Fix applied locally:** isolate each run's log file. `launchOnce` sets `MEMRY_TEST_LOG_DIR` to the run's fresh dir; `logger.ts` honors it via a test-gated `resolvePathFn` (worker-safe — `node:path` + `process.env`, no `electron` import); `logDir` is threaded through `LaunchedElectron`, and `waitForMainLog` reads only that isolated file (dropping the shared `getPath('logs')` candidate).

PR #903, opened 2026-07-27
fix(import): resolve markdown assets against the selected folder

h4yfans:
## Summary

User feedback: a Capacities Markdown export imports the notes fine but none of the images come across. Same failure shape applies to Obsidian-style vaults.

The attachment-copying feature already existed — `markdownImporter` extracts asset refs, saves them via `saveAttachment`, and rewrites the link to `memry-file://`. Two separate bugs meant it never ran for these exports:

**1. The traversal guard used the wrong reference point.** Refs were resolved against `path.dirname(note)` and anything escaping that directory was skipped as "path traversal". Capacities and Obsidian both keep media in a shared sibling folder and reference it as `../Images/Media/photo.png`, so every image was dropped — even though it sat well inside the folder the user selected.

Asset resolution is now bounded by the folder the user actually picked: `FileDescriptor` / `NotePlan` carry that root. A single picked file keeps its own directory as the root, so ad-hoc file imports are unchanged.

**2. Folder picking was unreachable from the UI.** The markdown `fileSpec` declared files only, so the native panel could never return a folder — contradicting both the shipped description ("Import .md files or a folder") and `apps/docs/src/user-guide/import.md`. The directory branch in the importer was only ever exercised by tests.

Electron cannot show one native panel that accepts both files and a folder — on Windows and Linux `['openFile', 'openDirectory']` silently degrades to directory-only. So importers that accept either now render a second button via a new `allowDirectory` flag. The existing `directory` flag still means directory-only (Apple Notes) and is untouched.

**3. A regression the widened root exposed.** With refs now resolvable across the selection, `[link](../other-note.md)` pointed at a real sibling note — and `.md` is in `ALLOWED_FILE_EXTENSIONS`, so notes were being copied in as *file attachments* and their links replaced with `<!-- file:{…} -->` blocks. Markdown refs are note-to-note links, not assets, and are now excluded (after stripping any `#anchor` / `?query`). `extractAssetRefs` feeds only this importer — Bear has its own local copy.

### Reviewer notes

- Widening the boundary from the note's directory to the selection root does not widen trust: the user explicitly picked that folder, which is the same consent model the Apple Notes directory pick already relies on. Single-file picks keep the old, narrower bound.
- The guard also rejects an absolute `path.relative` result, which is what you get on Windows when the two sides live on different drives.
- `apps/desktop/src/main/import/markdown/__fixtures__/nested-assets/` mirrors the reported export shape (notes in one folder, media in a sibling folder).

## Release note

Markdown import now brings images and attachments across when you pick a folder, including exports like Capacities and Obsidian that keep media in a shared folder alongside your notes.

## Test plan

- `pnpm test:desktop` — 12064 passed, 7 skipped, 0 failed
- `pnpm typecheck` — 16/16 packages
- `pnpm lint`, plus `eslint --no-cache --max-warnings=0` over the changed paths
- `pnpm ipc:check`, `pnpm check:architecture`, `pnpm check:contracts`
- `pnpm docs:impact --strict` and `pnpm docs:build`
- New coverage: an asset referenced above the note but inside the selected folder is copied; one that escapes the selected folder is still skipped; `extractAssetRefs` ignores `.md` / `.markdown` links including anchored ones
- Dry-run against the reported export: selecting the whole export resolves both refs (`copied=2 skipped=0`, including the 1002 KB PNG behind `../Images/Media/`); selecting only the notes subfolder still reports `skipped=1 (outside root)`

Not covered: Obsidian `![[image.png]]` embeds remain unsupported — `extractAssetRefs` drops wikilinks and the renderer draws them as a stray `!` plus a link pill. Capacities does not emit that form; tracked separately.


h4yfans:
Addressed the traversal-guard point from [this review](https://github.com/memrynote/memry/pull/903#pullrequestreview-4834622047); declining the i18n one.

**1. `startsWith('..')` false positive — fixed** (6459756a9)

Correct catch. `path.relative` returns `..media/shared.png` for a folder named `..media` sitting directly inside the selected root, and the guard read that as an escape and skipped every asset under it. Now only a whole parent segment is rejected:

```ts
const escapesRoot = refRelToRoot === '..' || refRelToRoot.startsWith(`..${path.sep}`)
if (escapesRoot || path.isAbsolute(refRelToRoot)) { … }
```

The cross-drive Windows case (`path.relative` returning an absolute path) is unchanged. New integration test builds a `..media/` sibling at runtime and asserts the asset is copied — verified it fails against the old guard (`attachments: expected 1, received 0`) and passes against the new one.

**2. English-only i18n keys — declining**

This is the repo's convention, not an oversight. `i18n:check` already reports 215 keys missing per locale and passes; every recent feature ships `en` only and the runtime falls back to English. Copying the English strings into the other 31 locales would render identically for users while marking the keys as translated, so a later translation pass would skip them.

**Verification**

- `pnpm --filter @memry/desktop test:main` — 4206 passed, 1 skipped, 0 failed
- `pnpm typecheck` — 16/16
- `eslint --no-cache --max-warnings=0` on both changed files
- `pnpm --filter @memry/desktop i18n:check`, `pnpm ipc:check`, `pnpm check:architecture`, `pnpm check:contracts`
- `pnpm docs:impact --base origin/main --strict` — covered

Separately, the symlink-escape note from [review 4834436351](https://github.com/memrynote/memry/pull/903#pullrequestreview-4834436351) is not addressed here: resolving through symlinks would need `fs.realpath` on both sides, and the boundary is a folder the user explicitly picked — same consent model the Apple Notes directory pick already relies on. Happy to tighten it if you'd rather not accept that.

PR #1113, opened 2026-08-07
perf(telemetry): journal queue enqueues instead of rewriting the mirror

### LINKED ISSUE #988: [MAJOR][telemetry] Every tracked event synchronously rewrites the entire queue file

> [!IMPORTANT]

---

**Problem.** `apps/desktop/src/main/telemetry/queue-store.ts:65-75` — `save()` does `fs.writeFileSync(filePath, JSON.stringify({version, items}))` with the **whole** queue. It is called from `apps/desktop/src/main/telemetry/client.ts:113-118` on every `trackMainEvent`/`trackMainLog`/`trackMainError` (204 call sites in main) and from `ship-queue.ts:63-68` for every `>= warn` log line. The queue drains only 50 items per 30 s flush (`TELEMETRY_BATCH_LIMIT = 50`, `TELEMETRY_QUEUE_LIMIT = 500`).

**Why it matters.** When the endpoint is unreachable (offline, dev builds), the queue pegs at 500 and every subsequent event serializes + sync-writes ~500 objects on the main thread. An error loop is worst-case: each error = one log-queue write + one event-queue write → quadratic degradation exactly when the app is unhealthy.

**Trigger.** Always; worst on error bursts / offline.

**Suggested fix.** Debounce the mirror write (e.g. 1–5 s trailing) and/or move it to async I/O; consider append-only journal + periodic compaction instead of full-file rewrite.

**Verification.** Unit test: N track calls within the debounce window produce 1 write. Manually: point telemetry at an unreachable host, generate 1k events, confirm no main-thread stalls (harness #446).

h4yfans:
Fixes #988

## Verification — the issue was still real

`apps/desktop/src/main/telemetry/queue-store.ts:65-75` on `main` @ `27bc2b96`:

```ts
save: (items) => {
  try {
    fs.writeFileSync(filePath, JSON.stringify({ version: FORMAT_VERSION, items }), 'utf-8')
```

`save(queue)` was called with the **whole** queue from `client.ts:117` (`track` → `persist()`) and
`ship-queue.ts:67` (`enqueue` → `persist()`) — i.e. on every `trackMainEvent` / `trackMainLog` /
`trackMainError` and every `>= warn` log line. With `TELEMETRY_QUEUE_LIMIT = 500` and a 50-item
drain per 30s flush, an unreachable endpoint pegs the queue at 500 and each further event
re-serialises and sync-writes all 500 objects on the main thread.

Measured on the branch by mutating the fix back to the old behaviour: a 100-event burst on a full
queue wrote **8,051,400 bytes** instead of **80,514** — a 100× amplification that grows with queue
depth.

## Change

The mirror is now an append-only journal instead of a full-file rewrite:

- `queue-store.ts` — on-disk format is a `{"version":2}` header line plus one JSON item per line.
  New `append(item, currentItems)` writes a single line via `fs.appendFileSync`; `save(items)` is
  unchanged in role (full rewrite) and is now the compaction path.
- `client.ts` / `ship-queue.ts` — only the enqueue path changes (`persist()` → `persistAppend()`).
  Drains, trims and opt-out still go through `save()` / `clear()`.

Why not debounce (the issue's first suggestion): the mirror exists so a hard crash cannot take
`app_crashed` with it. Deferring the write would reintroduce exactly the loss window the file was
added to close. The journal keeps **every enqueue on disk before it returns** and still makes the
write O(1) in queue depth.

Compaction: trimming drops from the head, which an append cannot express, so the journal is allowed
to carry trimmed-away lines until it passes `COMPACT_AFTER_LINES = 1000` (2× both queue limits), at
which point it is rewritten from the live queue. Callers already trim the restored set, so an
over-long journal is bounded on load as well.

### Backward compatibility

- **Upgrade:** `{"version":1,"items":[…]}` is still read, so an install upgrading from the shipped
  build keeps whatever its last session queued. The next write upgrades the file in place.
- **Downgrade:** the v1 reader `JSON.parse`s the whole file, fails on a multi-line journal, and takes
  its existing "unreadable mirror → warn, delete, treat as empty" path. Costs one session's queued
  telemetry, cannot wedge startup.
- Unknown versions are still discarded rather than parsed. A crash-truncated tail line is dropped
  rather than the whole journal.

## Tests

New `apps/desktop/src/main/telemetry/queue-store.test.ts` (7 tests) plus one test added to
`client.test.ts`. `ship-queue.test.ts` has one assertion updated from the raw v1 shape to the v2
shape (plus a behavioural restore check).

Note: `fs.appendFileSync` dispatches through `fs.writeFileSync` inside Node, so call counts cannot
separate the two paths — the tests assert **byte volume** and observable restore behaviour instead.

Red → green:

- Before the fix, `client.test.ts > does not rewrite the whole mirror once per tracked event`:
  `AssertionError: expected "writeFileSync" to be called +0 times, but got 100 times`
- After: `PASS (140) FAIL (0)` across `src/main/telemetry`.

Mutation-tested (reverted `persistAppend(event)` → `persist()` and dropped the compaction bound):

```
PASS (138) FAIL (2)
1. does not rewrite the whole mirror once per tracked event
   AssertionError: expected 8051400 to be less than 80514
2. compacts once the journal outgrows its bound
   AssertionError: expected [ +0, 1, 2, … ] to deeply equal [ Array(500) ]
```

Both went green again on restore, so neither test is vacuous.

## Checks

| Check | Result |
| --- | --- |
| `pnpm --filter @memry/desktop typecheck` | pass (contracts + architecture + ipc:check + node + web) |
| `pnpm lint` | pass — 0 errors, 14 pre-existing warnings, none in touched files |
| `vitest --project main` (full) | `PASS (5424) FAIL (0) skipped (4)` |
| `pnpm docs:impact --base origin/main --strict` | pass |
| `pnpm docs:build` | pass |

`apps/docs/src/architecture/observability.md` documented the old `{"version":1,"items":[…]}` format
verbatim, so it is updated in the same branch.

### DIFF
--- a/apps/desktop/src/main/telemetry/client.test.ts
+++ b/apps/desktop/src/main/telemetry/client.test.ts
@@ -407,6 +407,32 @@ describe('createTelemetryClient — crash durability', () => {
     expect(revived.getQueueDepth()).toBe(TELEMETRY_QUEUE_LIMIT)
   })
 
+  it('cannot be made to forge mirror entries by a hostile dimension value', () => {
+    // #given an event whose dimension carries remote-controlled text — inbox link
+    // metadata is scraped from arbitrary pages, which is the taint path CodeQL
+    // flags into this file — shaped to break out of a newline-delimited journal
+    const hostile = '\n{"id":"forged","name":"app_crashed","surface":"app"}\n'
+    const client = createTelemetryClient(createDeps({ persistPath }).deps)
+    client.track({
+      ...buildEvent('11111111-1111-1111-1111-111111111111', 'inbox_item_added'),
+      dimensions: { source: hostile }
+    })
+    client.track(buildEvent('22222222-2222-2222-2222-222222222222'))
+
+    // #then the payload was escaped rather than written raw: two events, two lines
+    const raw = fs.readFileSync(persistPath, 'utf-8')
+    expect(raw.split('\n')).toHaveLength(4) // header + 2 events + trailing ''
+
+    // #and the next launch restores the two real events, not a forged crash
+    const revived = createTelemetryClient(createDeps({ persistPath }).deps)
+    expect(revived.getQueueDepth()).toBe(2)
+    const restored = JSON.parse(
+      fs.readFileSync(persistPath, 'utf-8').split('\n')[1]
+    ) as TelemetryEvent
+    expect(restored.name).toBe('inbox_item_added')
+    expect(restored.dimensions?.source).toBe(hostile)
+  })
+
   it('never restores events for an install that opted out between launches', () => {
     // #given events mirrored while telemetry was on
     createTelemetryClient(createDeps({ persistPath }).deps).track(
--- a/apps/desktop/src/main/telemetry/client.test.ts
+++ b/apps/desktop/src/main/telemetry/client.test.ts
@@ -379,6 +379,34 @@ describe('createTelemetryClient — crash durability', () => {
     expect(createTelemetryClient(createDeps({ persistPath }).deps).getQueueDepth()).toBe(0)
   })
 
+  it('does not rewrite the whole mirror once per tracked event', () => {
+    // #given a client whose queue is already at the limit — what an unreachable
+    // endpoint produces, where every further event used to re-serialise all 500
+    const client = createTelemetryClient(createDeps({ persistPath }).deps)
+    for (let i = 0; i < TELEMETRY_QUEUE_LIMIT; i++) {
+      client.track(buildEvent(`11111111-1111-1111-1111-${String(i).padStart(12, '0')}`))
+    }
+    const mirrorSize = fs.statSync(persistPath).size
+    // fs.appendFileSync dispatches through fs.writeFileSync inside Node, so this
+    // spy sees both paths; byte volume is what separates them.
+    const write = vi.spyOn(fs, 'writeFileSync')
+
+    // #when 100 more events are tracked
+    for (let i = 1000; i < 1100; i++) {
+      client.track(buildEvent(`11111111-1111-1111-1111-${String(i).padStart(12, '0')}`))
+    }
+
+    // #then the whole burst wrote less than a single copy of the queue, where
+    // mirroring per event wrote 100 copies of it
+    const written = write.mock.calls.reduce((sum, call) => sum + String(call[1]).length, 0)
+    expect(written).toBeLessThan(mirrorSize)
+    write.mockRestore()
+
+    // #and a crash still loses nothing: the newest 500 events are on disk
+    const revived = createTelemetryClient(createDeps({ persistPath }).deps)
+    expect(revived.getQueueDepth()).toBe(TELEMETRY_QUEUE_LIMIT)
+  })
+
   it('never restores events for an install that opted out between launches', () => {
     // #given events mirrored while telemetry was on
     createTelemetryClient(createDeps({ persistPath }).deps).track(
--- a/apps/desktop/src/main/telemetry/client.ts
+++ b/apps/desktop/src/main/telemetry/client.ts
@@ -87,6 +87,12 @@ export const createTelemetryClient = (deps: TelemetryClientDeps): TelemetryClien
     store?.save(queue)
   }
 
+  // The enqueue path appends one line rather than rewriting the mirror, so a
+  // tracked event costs the same whether the queue holds 1 event or 500.
+  const persistAppend = (event: TelemetryEvent): void => {
+    store?.append(event, queue)
+  }
+
   // A mirror written by a build with a larger limit must not resurrect an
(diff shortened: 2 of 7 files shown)


h4yfans:
## CodeQL alert 129 — `js/http-to-file-access`, verified and locked with a test

**Alert:** "Network data written to file — Write to file system depends on Untrusted data", medium, at `queue-store.ts:143` (the `fs.appendFileSync` in `append`). Not dismissed — leaving that call to @h4yfans.

### The taint source is real, so this was worth checking

The alert names its source: `apps/desktop/src/main/inbox/metadata.ts#L91`, which is `net.fetch` inside `chromiumFetch` — the metascraper fetch that pulls title/description/author off **arbitrary URLs the user clips**. That text can reach a telemetry dimension, so "untrusted data reaches this write" is accurate, not a false path.

Two separate questions, though:

**1. Path traversal / arbitrary file upload — not applicable.** `filePath` is fixed at construction from `app.getPath('userData')` (`index.ts:1413`/`1418`); no tainted value reaches the path argument, only the data argument. And the same data already reached the same file before this PR — that part is not new.

**2. Line forging — this IS new, and is the right thing to check.** This PR made the format newline-delimited, so a raw `\n` in a value would forge journal entries or truncate the parse. Your read is correct.

### Verification, in the code

Every byte of the journal is one of exactly two expressions, both in `queue-store.ts`:

```ts
const encodeItem = (item: T): string => `${JSON.stringify(item)}\n`
// save():
fs.writeFileSync(filePath, header + items.map(encodeItem).join(''), 'utf-8')
// append():
fs.appendFileSync(filePath, encodeItem(item), 'utf-8')
```

with `header = ${JSON.stringify({ version: FORMAT_VERSION })}\n` (constant). **No part of any line is assembled by concatenating a value** — the only non-constant text is `JSON.stringify` output, and `JSON.stringify` escapes all code units below `U+0020` (including `U+000A`/`U+000D`) inside **both keys and values**. No `space` argument is passed, so no pretty-print newlines either.

Two details worth recording:

- `U+2028`/`U+2029` are emitted **raw** by `JSON.stringify` (they are legal in JSON strings). They are harmless here only because `parse()` splits on `\n` alone — that is now asserted, so a future switch to a line splitter that treats them as breaks fails the test.
- `JSON.stringify` throws on cyclic input and BigInt. Both call sites already have it inside the `try`, so it takes the existing `onWriteFailure` path rather than escaping.

### The test that locks it

`queue-store.test.ts` — "cannot be made to forge journal lines by an item carrying newlines": items whose values (and one **key**) carry `\n`, `\r\n`, lone `\r`, `\t`, `NUL`, `U+001F`, `U+2028`/`U+2029`, quotes/backslashes, plus text shaped like a complete forged entry (`\n{"id":"forged","name":"app_crashed"}\n`) and a forged header (`\n{"version":2}\n`). Asserts exact line count and byte-for-byte restore.

`client.test.ts` — "cannot be made to forge mirror entries by a hostile dimension value": same payload through the **real taint shape**, a `TelemetryEvent` with `dimensions: { source: <hostile> }`, tracked via `client.track()`.

**Red → green.** Mutated `encodeItem` to unescape newlines after stringify (simulating a refactor that concatenates):

```
PASS (144) FAIL (2)
1. client — cannot be made to forge mirror entries by a hostile dimension value
   AssertionError: expected [ '{"version":2}', …(5) ] to have a length of 4 but got 6
2. queue-store — cannot be made to forge journal lines by an item carrying newlines
   AssertionError: expected [ '{"version":2}', '{"lf":"a', …(19) ] to have a length of 5 but got 21
```

6 lines instead of 4 in the client case is the forged `app_crashed` entry actually materialising. Restored → `PASS (146) FAIL (0)`.

### Truncated tail (point 3)

Already covered before this push: `queue-store.test.ts` — **"drops a line the crash truncated instead of the whole journal"** writes `{"a":` after a complete journal and asserts the preceding items survive and only the partial line is lost.

### Codecov

The uncovered line was `onWriteFailure(error)` in `append`'s catch (`queue-store.ts:147`) — a **real** branch: the existing "cannot be written" test goes through `save()`, because `journalStarted` is false on first call, so the append-path failure had never executed. Covered with a mid-session `ENOSPC` (journal already started).

Also covered three genuinely reachable corrupt-file branches that istanbul flagged as partial: a non-object header, a v1 header whose `items` is not an array, and a v2 header with no items yet (crash between header and first append).

`queue-store.ts` is now **67/67 statements, 0 uncovered branches**. Nothing left that I would call unreachable defensive code.

### Checks re-run

| Check | Result |
| --- | --- |
| `pnpm --filter @memry/desktop typecheck` | pass |
| `pnpm lint` | pass — 0 errors, 14 pre-existing warnings, none in touched files |
| `vitest --project main src/main/telemetry` | `PASS (146) FAIL (0)` |
| `vitest --project main` (full) | `PASS (5430) FAIL (0) skipped (4)` |

Still draft, not merged.

PR #1138, opened 2026-08-07
fix(viewers): load the PDF once and window the thumbnail rail

### LINKED ISSUE #1018: [MAJOR][viewers] PDF viewer loads the document twice and renders every page as an unvirtualized thumbnail

> [!IMPORTANT]

---

**Problem.** `apps/desktop/src/renderer/src/components/viewers/pdf-viewer.tsx:229-263,266-283` — two separate `<Document file={src}>` instances (sidebar + main) = two independent pdf.js `PDFDocumentProxy` objects, each with its own worker-side document and page cache. The sidebar mounts one `<Page>` per page with no windowing: a 400-page PDF mounts 400 canvases (~120×160×4 B ≈ 77 kB each → ~30 MB backing store) and kicks off 400 concurrent render tasks on open. No explicit `pdf.destroy()` or render-task cancellation (relies on react-pdf internals).

**Trigger.** Opening any long PDF.

**Suggested fix.** Share one `<Document>` (or one loaded proxy) between main view and sidebar; virtualize the thumbnail rail (e.g. `@tanstack/react-virtual`, already used in calendar); verify document teardown with a heap snapshot after 10 open/close cycles.

**Verification.** 400-page PDF opens with O(visible) thumbnail canvases; single worker document; memory returns to baseline after closing the tab.

h4yfans:
Fixes #1018

## Verification — the issue was still real

`apps/desktop/src/renderer/src/components/viewers/pdf-viewer.tsx` on `origin/main` (27bc2b961):

- **Two documents.** `<Document file={src}>` at line 234 (sidebar) and line 268 (main pane). Two mounts of react-pdf's `Document` means two `pdfjs.getDocument()` loading tasks and two independent `PDFDocumentProxy` objects for the same file (`react-pdf/dist/Document.js` → `loadDocument`).
- **No windowing.** Line 235: `Array.from({ length: numPages }, ...)` mounted one `<Page width={120}>` per page inside the sidebar `Document`, so a 400-page PDF mounted 400 canvases and queued 400 render tasks on open.

The characterisation test reproduced both against unmodified `main`:

```
× loads the file into a single pdf.js document for both surfaces
  → expected [ <div …(2)>…(400)</div>, …(1) ] to have a length of 1 but got 2
× windows the thumbnail rail instead of mounting one canvas per page
  → expected 400 to be less than 20
× releases the document proxy and every page canvas on unmount
  → expected [ { …(2) }, { …(2) } ] to have a length of 1 but got 2
```

## Change

Single file changed: `apps/desktop/src/renderer/src/components/viewers/pdf-viewer.tsx`.

1. **One document.** The single `<Document>` now wraps the whole content row, so the rail and the page view share one proxy via react-pdf's `DocumentContext`. The local `loading` state existed only to gate the second `<Document>` and is gone — react-pdf's own `loading` prop already covers the pane.
2. **Windowed rail.** New `PdfThumbnailRail` renders only the rows in view using `@tanstack/react-virtual`, **already a `@memry/desktop` dependency** (`apps/desktop/package.json:193`, used by the task lists, notes tree and folder table). No new dependency, no lockfile change.
   - Rows are a fixed height (`156px` preview + label), so the window does not shift while pdf.js renders asynchronously.
   - The rail scrolls natively instead of through `ScrollArea`: the virtualizer needs a ref to the element that actually scrolls, and Radix keeps that inside its own viewport.
   - A `scrollToIndex(currentPage - 1, { align: 'auto' })` effect pulls the active thumbnail back into view when the page changes from the toolbar, since it may be outside the window.
   - The active row carries `aria-current="page"` in addition to the existing highlight.

**Large documents:** no page cap and no lazy `numPages`. A 500-page PDF still opens fully — the toolbar reports `1 / 500`, the rail's scroll range covers all 500 rows, and any page is reachable by scrolling or by the page controls. Only the ~4 rows in the viewport (plus 2 overscan each side) hold a canvas at a time, versus 500 before.

Behaviour deliberately unchanged: toolbar, zoom, rotate, fit-to-width, error state, `hidden sm:block` rail visibility, 140px rail width, 120px thumbnails.

## Tests

New `apps/desktop/src/renderer/src/components/viewers/pdf-viewer.test.tsx` (5 tests, 400-page document). The `react-pdf` mock mirrors the real teardown contract — `Document` registers one loading task per mount and destroys it in its effect cleanup; `Page` holds a canvas while mounted. `@tanstack/react-virtual` is **not** mocked; the test gives the rail a real `offsetHeight` and drives real scroll events.

- single `<Document>` / single loaded proxy for both surfaces
- fewer than 20 thumbnail canvases mounted for 400 pages, and live-canvas count matches the DOM
- scrolling the rail to page 300 renders that region (and drops page 1), clicking it navigates to `300 / 400` and marks it `aria-current`
- after scrolling the rail to page ~350, the toolbar's next-page control makes the rail scroll back so page 2 is rendered again
- unmount destroys the document proxy and leaves zero live canvases

Red → green: all 5 failed on unmodified `main` (output above), all 5 pass after the change. The existing `viewers.test.tsx` (5 tests) still passes untouched.

Mutation-tested (the fix reverted one line at a time, test must go red):

- drop `virtualizer.scrollToIndex(...)` → `pulls the active thumbnail back into view` fails (`expected "vi.fn()" to be called at least once`)
- `overscan: 2` → `overscan: 500` → `expected 400 to be less than 20`, plus 2 more

## Checks

| check | result |
| --- | --- |
| `pnpm --filter @memry/desktop typecheck` | pass (contracts + architecture + ipc:check + node + web) |
| `pnpm lint` | pass — 0 errors, 17 warnings (14 pre-existing) |
| `pnpm exec vitest run --project renderer .../viewers/` | 10 passed (2 files) |
| `npx react-doctor@latest .` | no findings in `pdf-viewer.tsx`; score unchanged at 64 (repo-wide pre-existing) |
| `pnpm docs:impact --base origin/main --strict` | `covered` |
| `pnpm docs:build` | pass |

The 3 new lint warnings are all on the `useVirtualizer` call and its follow-scroll effect: one `react-hooks/incompatible-library` ("TanStack Virtual's `useVirtualizer()` API returns functions that cannot be memoized safely" — inherent to the library) and two `react-you-might-not-need-an-effect/no-pass-ref-to-parent` false positives on the `scrollToIndex` effect. All warnings, not errors; the lint gate does not set `--max-warnings`.

## Docs

`apps/docs/src/user-guide/notes/attachments.md` — one sentence under **PDF Inline Preview** noting the file-page thumbnail sidebar draws thumbnails on demand. Needed to clear the `missing-docs` gate, and it is accurate user-visible behaviour.

### DIFF
--- a/apps/desktop/src/renderer/src/components/viewers/pdf-viewer.test.tsx
+++ b/apps/desktop/src/renderer/src/components/viewers/pdf-viewer.test.tsx
@@ -1,0 +1,232 @@
+/**
+ * Regression tests for the PDF viewer's document/thumbnail memory footprint.
+ *
+ * Covers two defects reported in issue #1018:
+ *  1. the viewer mounted one `<Document>` per surface, so a single file was
+ *     loaded twice into two independent pdf.js document proxies;
+ *  2. the thumbnail rail mounted one `<Page>` canvas per page with no
+ *     windowing, so a long PDF allocated hundreds of canvases on open.
+ *
+ * The `react-pdf` mock mirrors the real teardown contract: `Document` creates
+ * one loading task per mount and destroys it in its effect cleanup (see
+ * `react-pdf/dist/Document.js` -> `loadDocument`), and `Page` holds a canvas
+ * for as long as it stays mounted.
+ */
+import { fireEvent, render, screen } from '@testing-library/react'
+import userEvent from '@testing-library/user-event'
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
+
+import { PdfViewer } from './pdf-viewer'
+
+const NUM_PAGES = 400
+const RAIL_VIEWPORT_HEIGHT = 600
+
+const pdfMocks = vi.hoisted(() => ({
+  loadedDocuments: [] as { file: string; destroyed: boolean }[],
+  liveCanvases: new Set<{ page: number; kind: 'main' | 'thumbnail' }>()
+}))
+
+vi.mock('@memry/i18n/renderer', () => ({
+  useT: () => ({ t: (key: string) => key })
+}))
+
+vi.mock('react-i18next', () => ({
+  getI18n: () => ({ getFixedT: () => (key: string) => key })
+}))
+
+vi.mock('react-pdf', async () => {
+  const { useEffect, useState } = await import('react')
+
+  return {
+    pdfjs: { GlobalWorkerOptions: {} },
+    Document: ({
+      file,
+      children,
+      className,
+      loading,
+      onLoadSuccess
+    }: {
+      file: string
+      children?: React.ReactNode
+      className?: string
+      loading?: React.ReactNode
+      onLoadSuccess?: (result: { numPages: number }) => void
+    }) => {
+      const [ready, setReady] = useState(false)
+
+      useEffect(() => {
+        const proxy = { file, destroyed: false }
+        pdfMocks.loadedDocuments.push(proxy)
+        setReady(true)
+        onLoadSuccess?.({ numPages: NUM_PAGES })
+        return () => {
+          proxy.destroyed = true
+        }
+      }, [file, onLoadSuccess])
+
+      return (
+        <div data-testid="pdf-document" data-file={file} className={className}>
+          {ready ? children : loading}
+        </div>
+      )
+    },
+    Page: ({ pageNumber, width }: { pageNumber: number; width?: number }) => {
+      useEffect(() => {
+        const canvas = {
+          page: pageNumber,
+          kind: (width ? 'thumbnail' : 'main') as 'main' | 'thumbnail'
+        }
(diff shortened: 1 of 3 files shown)


h4yfans:
Review note for whoever merges this — one visual change that no automated test covers.

The thumbnail rail is now windowed with a **fixed 182px row pitch** (156px preview + label). That fixed height is load-bearing: without it the window shifts while pdf.js renders pages asynchronously, so it is the right call for virtualization. The consequence is that an unusually tall or landscape page is centered inside the 156px box instead of sizing its own row, which is a small change from the previous behaviour where every thumbnail sized itself.

Worth opening a landscape or mixed-orientation PDF by hand before merging. Not filing this as an issue since it is a property of this PR, not of main.

PR #1166, opened 2026-08-07
fix(inbox): enqueue a sync push when an item is filed

### LINKED ISSUE #1159: [HIGH][inbox][sync] Filing an inbox item never enqueues a sync push — filed items reappear in the Inbox on every device
## Where this came from

Reported by a user over email, 2026-08-05 — two reports 17 minutes apart, on `v2026-07-19.2`:

> I'm not sure if this is by design, but when using the webclipper on my MacBook I "filed" a few sites to move them out of the Inbox. After syncing on my Fedora laptop the web clips synced properly, but they still show up in the Inbox. Will Inbox status sync?

> Oddly, when I went back to my MacBook the web clipped notes were back in the inbox after filing earlier.

Not by design. Both halves are the same defect.

## The defect

`markItemAsFiled()` — `apps/desktop/src/main/inbox/filing.ts:448` — writes `filedAt` / `filedTo` / `filedAction`, publishes an `inbox.upserted` projection event, emits the `FILED` IPC event, and **never enqueues a sync push**. `filing.ts` does not import `syncInboxUpdate` at all.

Every filing path funnels through it — 8 call sites in `filing.ts` (folder, note-folder, note, task, event, reminder, linked ×2). So no form of filing reaches another device.

The contrast makes it clearly an omission, not a decision:

| Action | Enqueues push? |
|---|---|
| File (all 8 paths) — `filing.ts:448` | **no** |
| Undo-file — `crud.ts:283` | yes |
| Snooze / unsnooze / auto-unsnooze — `snooze.ts:215, 273, 406` | yes |
| Create / update / archive / delete — `crud.ts:81, 105, 191, 223, 315` | yes |

**Unfiling syncs. Filing does not.**

### Why the item comes *back* on the device that filed it

`buildPushPayload` (`inbox-handler.ts:146`) ships the whole row. A peer that never learned the item was filed still holds `filedAt: null`, and pushes that row on its own next change to the item.

`applyUpsert` (`inbox-handler.ts:56`) deliberately distinguishes "key omitted" from "key present and null" via `hasKey` — an explicit null **is** the remote unfile clear. The stale peer's payload has `filedAt` present and null, so the origin device treats it as a deliberate unfile and wipes its own filed state.

That is the second report: filed on the MacBook, came back on the MacBook.

## Why it matters

Production app, real user data, live beta. The Inbox is the triage surface and filing is its primary action — on a multi-device setup the Inbox never drains, and work the user already did silently undoes itself. It reads as the app losing their input.

## What to confirm first

Verified against `main` on 2026-08-08; line numbers may drift.

Note that PR #955 (`998e0cc6f`, 2026-08-05) hardened the **apply** side of this handler — carrying `filedAt`/`filedTo`/`filedAction` on insert so a first-time peer doesn't resurrect filed items, plus the `hasKey` semantics above. It did not touch the **push** side, which is where this lives. The reporting build, `v2026-07-19.2`, predates even that.

## Possible directions (not a decision)

- Call `syncInboxUpdate(itemId)` from inside `markItemAsFiled` — one site, covers all 8 paths. Wrap it the way `convertToTask` does after #973, so a sync-enqueue throw doesn't fail an otherwise successful filing.
- Consider whether the projection publish, the IPC emit and the sync enqueue should be a single helper, so the next state change added to `inboxItems` can't skip the queue the same way.
- A guard test worth having: every write to `inboxItems` under `src/main/inbox/**` either enqueues a push or is explicitly listed as local-only.

No payload or protocol change is needed for the fix itself, so there is no compat risk beyond the usual — but see `CLAUDE.md`: older builds must keep working.

h4yfans:
Fixes #1159

## Summary

`markItemAsFiled()` wrote `filedAt` / `filedTo` / `filedAction`, published the projection event and emitted the `FILED` IPC event — but never enqueued a sync push. `filing.ts` did not import `syncInboxUpdate` at all. All 8 filing paths funnel through that one function, so **no form of filing reached another device**, while undo-file, snooze, archive, create, update and delete all pushed.

The reporter's second symptom follows from the same omission. `buildPushPayload` ships the whole row, so a peer that never learned the item was filed still holds `filedAt: null` and pushes that row on its own next change to the item. `applyUpsert` reads a present-and-null `filedAt` as the deliberate remote unfile clear (the `hasKey` semantics from #955), so the origin device wipes its own filed state — the item reappears on the very device that filed it.

**The fix:** route the write through `syncInboxUpdate(itemId)`, which enqueues the push *and* publishes the same projection. It also fires `emitCalendarProjectionChanged` / `scheduleGoogleCalendarSourceSync`, which filing genuinely wants since it clears `snoozedUntil` — a second, quieter gap that closes for free.

Best-effort like `convertToTask`'s enqueue (#973): a sync outage must not fail a filing that already succeeded locally. One addition — the catch republishes via `publishInboxUpserted`, because a throw inside `syncInboxUpdate` happens *before* its own projection publish and would otherwise leave the local Inbox list stale.

## Compat

No payload, schema or protocol change. An older peer just starts receiving filed rows it already knows how to apply — #955 hardened that apply side. Nothing new is required of older builds.

## Guard test

`src/main/inbox/sync-enqueue-guard.test.ts` scans every non-test file under `src/main/inbox/`, slices out `update(inboxItems)` / `insert(inboxItems)` statements up to `.run()`, and fails any file that writes a **triage column** (`filedAt`, `filedTo`, `filedAction`, `archivedAt`, `snoozedUntil`, `snoozeReason`) without referencing `syncInbox(Create|Update|Delete)`. Statement slicing matters — those column names appear in plenty of selects and filters.

Mutation-verified: strip the fix and it reports `['filing.ts']`.

### Why triage columns and not every write

The issue floated "every write to `inboxItems` either enqueues a push or is explicitly listed as local-only". `capture.ts`, `jobs.ts`, `ingest.ts` and `transcription.ts` all write `inboxItems` without enqueuing — allowlisting them would assert they are local-only, which is false. They are a real but milder gap (enrichment rides along on the row's next syncing write; only a never-touched-again item stays diverged), tracked separately rather than papered over here.

## Release note

Filing an item in the Inbox now syncs to your other devices. Previously a filed item stayed filed only on the device you filed it from, and could reappear in the Inbox there once another device synced.

## Test plan

- 9 new tests in `filing.test.ts`, one per filing path (folder text + binary, note, task, event, reminder, link text + binary) plus an enqueue-throws resilience case asserting the filing persists, the projection is still published and the failure is counted. Verified RED first — 9/9 failed before the fix.
- `pnpm exec vitest run --config config/vitest.config.ts --project main` — 5452 passed, 0 failed, 4 skipped
- `pnpm lint` — 0 errors (15 pre-existing warnings)
- `pnpm --filter @memry/desktop typecheck:node`, `pnpm check:architecture`, `pnpm check:contracts` — pass
- `pnpm docs:impact --base origin/main --strict` — covered; `pnpm docs:build` — pass
- `git diff --check` — clean

### DIFF
--- a/apps/desktop/src/main/inbox/filing.test.ts
+++ b/apps/desktop/src/main/inbox/filing.test.ts
@@ -29,6 +29,16 @@ vi.mock('../database/queries/notes', () => ({
   setNoteTags: (...args: unknown[]) => mockSetNoteTags(...args)
 }))
 
+const { mockSyncInboxUpdate, mockPublishInboxUpserted } = vi.hoisted(() => ({
+  mockSyncInboxUpdate: vi.fn(),
+  mockPublishInboxUpserted: vi.fn()
+}))
+
+vi.mock('./runtime-effects', () => ({
+  syncInboxUpdate: mockSyncInboxUpdate,
+  publishInboxUpserted: mockPublishInboxUpserted
+}))
+
 const mockSend = vi.fn()
 const mockRename = vi.fn()
 const mockCopyFile = vi.fn()
@@ -191,6 +201,8 @@ describe('Inbox Filing Operations', () => {
     mockSetTaskTags.mockReset()
     mockGetInboxProject.mockReset().mockReturnValue({ id: 'project-inbox' })
     mockCreateReminder.mockReset().mockReturnValue({ id: 'rem_1' })
+    mockSyncInboxUpdate.mockReset()
+    mockPublishInboxUpserted.mockReset()
     vi.mocked(getStatus).mockReturnValue({ isOpen: true, path: '/mock-vault' } as never)
 
     mockCreateNote.mockResolvedValue({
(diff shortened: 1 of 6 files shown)


h4yfans:
## Added: the backfill for installs that already filed

The original fix makes new filings sync. It does nothing for items already filed by a pre-fix build, and I could not find any path that would ever heal them:

- `sync/dirty-recovery.ts` enumerated task / project / note / journal — inbox was absent.
- `inbox-handler.seedUnclocked` only picks rows with a **NULL** clock. These rows have one; they were pushed at capture time.
- `sync/manifest-check.ts` is presence-based (`!serverRef`), and an already-filed item **is** present on the server.
- The inbox clock bump lives in `inbox-sync.ts` `applyLocalChange`, reachable only through the enqueue path — so a pre-fix filing never advanced the clock either.
- `filing.ts` rejects re-filing an item that already has a `filedAt`, so nothing ever re-touches the row.

Net effect: every item filed by an older build stayed unpushed forever, and stayed exposed to the stale-peer push that reads its `filedAt: null` as a deliberate unfile.

### Predicate

New `recoverDirtyInbox` arm in `dirty-recovery.ts`, same shape as the note/journal arms:

```
clock IS NOT NULL
AND local_only IS NOT 1
AND (synced_at IS NULL OR modified_at > synced_at)
```

`modifiedAt > syncedAt` is what actually distinguishes a stranded row — `markItemAsFiled` does stamp `modifiedAt`, and the failed push left `syncedAt` behind. A NULL-clock predicate would have been wrong: these rows are clocked, that is precisely why `seedUnclocked` misses them.

Cost is one full-table read per sync-runtime start that returns **zero rows on a healthy install**, not a re-push of the inbox. It is self-clearing — the push stamps `syncedAt` past `modifiedAt` via `inboxHandler.markPushSynced`, so the next launch finds nothing. Before it drains, `SyncQueueManager.enqueue()` deduplicates on itemId+type+operation, so repeated launches cannot pile up rows. Both properties are covered by a test.

It deliberately uses the ordinary `enqueueUpdate`, which **bumps the vector clock**, rather than `enqueueRecoveredUpdate` like the other arms. `enqueueRecoveredUpdate` re-pushes a change whose clock already advanced at write time; these rows never got that far, so replaying their stored clock would lose to any peer that has moved on since (`compare` → `after` → `skip`) and the filing would be dropped a second time. Bumping produces exactly the push `markItemAsFiled` would have produced at filing time.

One deliberate side effect worth naming: enrichment-only writes that touch `modifiedAt` without enqueueing are now swept too. That is real local divergence, and pushing it is a fix, not a regression — but it is a behaviour change beyond filing.

### Guard tightening

`sync-enqueue-guard.test.ts` ran `SYNC_ENQUEUE_PATTERN.test(source)` against the **whole file**. Post-fix `filing.ts` contains `syncInboxUpdate`, so any *additional* triage write added to `filing.ts` / `crud.ts` / `snooze.ts` later would have passed vacuously — only a brand-new file with zero enqueue references could be caught. Three changes:

1. **Per-write, not per-file.** Each `update(inboxItems)` / `insert(inboxItems)` statement that assigns a triage column is matched against its own brace-matched enclosing block.
2. **String literals and comments stripped** before the match. `markItemAsFiled`'s catch logs `'syncInboxUpdate failed; filing persisted locally'`, and that mention alone satisfied the old check — I hit this while mutation-testing.
3. **Scope widened to the whole main process.** `readdirSync(INBOX_DIR)` was non-recursive and directory-scoped, so `calendar/google/sync-service.ts` — which clears `snoozedUntil` on `inbox_items` — sat outside the fence entirely. It is now scanned and passes, via `publishInboxCalendarMutation` → `tryEnqueueProjectionSyncUpdate`, which the pattern now recognises.

Genuinely local-only writers are allowlisted by path with a stated reason. Today that is only `sync/item-handlers/inbox-handler.ts`, whose writes *are* the incoming peer push — enqueueing there would echo every pulled change back to the server. A test fails if an allowlist entry goes stale, and the tripwire test asserts the scan still reaches all five triage writers.

### Verification

Rebased onto current `origin/main` first (was 209 behind) — clean, no conflicts.

- `pnpm lint` — 0 errors, 92 warnings, all pre-existing and none in the changed files
- `pnpm typecheck` — 16/16 tasks successful
- `npx vitest run --config config/vitest.config.ts --project main src/main/sync/dirty-recovery.test.ts src/main/inbox/sync-enqueue-guard.test.ts src/main/sync/note-title-sync-regression.test.ts src/main/inbox` — **28 files, 402 tests passed**
- `pnpm docs:impact --base origin/main --strict` — covered · `pnpm docs:build` — build complete
- `git diff --check` — clean

**Mutation-tested, both halves:**

- Replaced `recoverDirtyInbox(db, adapters)` with `0` → the 3 new inbox recovery tests failed (`expected +0 to be 1`), the rest stayed green. Restored → green.
- Appended a new unguarded `db.update(inboxItems).set({ filedAt: null })` to `filing.ts` — the exact case the file-level guard let through → the guard failed with `expected [ 'inbox/filing.ts:1285 [filedAt]' ] to deeply equal []`. Restored → green.

The forward-path fix is untouched.

Issue #987, opened 2026-08-07
[EPIC] Memory & CPU: full-app performance audit 2026-08 — fix tracking


h4yfans:
## Context

A production user (20 GB Mac) reported excessive RAM usage. The app is designed to stay open all day with every feature in use (inbox, journal, calendar, voice recorder, Agent Chat, graph, canvas, sync). A full static audit of the desktop app was performed on 2026-08-07 (`main` @ e4c9c90ee) across 8 subsystems: main-process lifecycle, sync/CRDT, database/indexer/projections, renderer hooks/services, heavy components (graph/editor/canvas/PDF), Agent Chat/MCP, media/voice/notifications, and the IPC boundary.

Every finding below is filed as its own issue, all carrying the **`audit-2026-08`** label (plus `memory-footprint`). Severity is in each title: `[MAJOR]` = unbounded growth under normal daily use or a permanent CPU burn · `[HIGH]` = grows/leaks on a common action · `[MEDIUM]` = bounded but wasteful · `[LOW]` = minor/latent · `[CLEANUP]` = dead code carrying latent cost.

## Rules for implementers (AI agents included)

Every child issue starts with the same instruction: **first confirm the issue is still valid on current `main`** (line numbers may have drifted; a later commit may have fixed it). Only implement after confirming. If invalid, comment evidence and close. Follow `CLAUDE.md` conventions, keep changes surgical, add tests, and use the memory snapshot harness (#446) for before/after evidence where applicable.

## Most likely explanation of the 20 GB report

Y.Docs pinned forever by stale `windowIds` (compaction disabled as a side effect) · editors never destroyed + journal remounting an editor per external sync change · agent transcripts never evicted · orphaned Whisper utility processes (hundreds of MB each) · leaked PDF-export renderer processes · full note bodies retained in `debugState` · offline CRDT buffers.

## Related pre-existing issues (folded into this epic)

- #449 — sync progress/conflict pruning (duplicates audit finding; already filed)
- #448 — unbounded renderer caches incl. URL previews (duplicates audit finding; already filed)
- #450, #451, #447, #445 — partial overlaps, referenced from the corresponding audit issues
- #446 — memory snapshot harness (verification infrastructure for all of the above)

### 🔴 MAJOR — unbounded growth in daily use / permanent CPU burn (32)
- [x] #988 — [telemetry] Every tracked event synchronously rewrites the entire queue file
[...]

PR #1213, opened 2026-08-11
perf(settings): make the settings:changed echo cost zero re-renders

### LINKED ISSUE #1063: [MEDIUM][settings] Every settings write echoes settings:changed back to the originating window (7 sites; sync apply sends 3 per window)

> [!IMPORTANT]

---

**Problem.** `apps/desktop/src/main/ipc/settings-handlers.ts:345,389,478,524,794,837,989` — all fan out to `getAllWindows()` with no `event.sender` exclusion, so the writer re-receives its own change and re-renders every `onSettingsChanged` subscriber (a dozen hooks). `sync/item-handlers/settings-handler.ts:163-194` sends up to **three** separate CHANGED messages per window in one apply.

**Suggested fix.** Exclude `event.sender` from the fan-out (renderer already applied its change optimistically) or include a `sourceWindowId` the renderer can filter; coalesce the sync-apply trio into one message.

**Verification.** Changing a setting does not re-fire the originating window's subscribers; sync apply emits one message per window.

h4yfans:
Closes #1063. Part of epic #987 (memory & CPU audit 2026-08).

## What was wrong

Every settings write fans `settings:changed` out to all windows, including the one that issued the write:

- `apps/desktop/src/main/ipc/settings-handlers.ts:346, 388, 475, 519, 787, 828, 978` — 7 write sites, all `broadcastToAllWindows(...)`.
- `apps/desktop/src/main/sync/item-handlers/settings-handler.ts:168-193` — sends up to **three** separate CHANGED messages per apply, and each one re-broadcasts the *whole* merged `general` / `editor` / `inbox` group, not just the fields that differ.

Each of the ~13 renderer subscribers merged the payload with `setSettings((prev) => ({ ...prev, ...value }))`. That spread always mints a new object, so an echo carrying **nothing new** still produced a state update and re-rendered every consumer. Every sync apply re-rendered every settings consumer in every window even when no field on this device had actually changed.

## Why the suggested fix was not safe

The issue suggested excluding `event.sender` from the fan-out, or filtering on a `sourceWindowId` in the renderer. Both break convergence.

A single window holds **many independent instances of the same settings hook** — each call site has its own `useState`, with no shared context:

| hook | consumer files |
| --- | --- |
| `useGeneralSettings` | 28 |
| `useFeatureFlags` | 9 |
| `useEditorSettings` | 3 |

Only the instance that called `updateSettings` applies the change optimistically. The other 27 instances of `useGeneralSettings` in that same window converge **only** through the echo. Dropping it at the sender (or filtering it per window in the renderer — same granularity) would leave e.g. the sidebar, journal panel and note header stale after changing the theme in Settings, until reload. This is the opposite of the `crdt-provider` case, where skipping the source window is correct.

So the dedupe belongs at the granularity that is actually correct: **the value**.

## What changed

New `apps/desktop/src/renderer/src/lib/settings-patch.ts`:

```ts
export function mergeSettingsPatch<T extends object>(prev: T, patch: Partial<T>): T
```

Returns `prev` **by identity** when every patched key already matches. React then bails out of the state update without rendering children or firing effects, so a no-op echo is free. Applied to all 10 group hooks (`use-general-settings`, `use-editor-settings`, `use-task-preferences`, `use-feature-flags`, `use-journal-settings`, `use-tab-preferences`, `use-inbox-preferences`, `use-keyboard-settings`, `use-calendar-preferences`, `use-note-editor-settings`) plus the tabs `UPDATE_SETTINGS` reducer, which now returns the existing state object unchanged.

No IPC contract, payload shape, settings shape or main-process behaviour changed — renderer-side only, so there is nothing for existing installs to migrate.

Measured React behaviour (probed, not assumed): React cannot use its eager-state shortcut on the first dispatch after a real state change, so that one still renders the hook's own component once before bailing out. Every subsequent no-op echo costs **zero** renders — which is the steady state, since each sync apply re-broadcasts the unchanged group.

## Verification

Mutation-tested: reverting `mergeSettingsPatch` to an unconditional `{ ...prev, ...patch }` fails **9** of the new tests, restoring it passes all 28.

```
Test Files  3 failed (3)      <- with the fix reverted
     Tests  9 failed | 19 passed (28)

Test Files  3 passed (3)      <- with the fix
     Tests  28 passed (28)
```

Wider suites covering everything touched (tabs contexts + reducers, feature flags):

```
Test Files  14 passed (14)
     Tests  134 passed (134)
```

- `pnpm --filter @memry/desktop typecheck:web` — exit 0
- `pnpm exec eslint <changed files>` — 0 errors, 0 warnings
- `git diff --check` — clean
- `pnpm docs:impact --base origin/main --strict` — exit 0
- `pnpm docs:build` — build complete

## Deliberately not done

- **Coalescing the sync-apply trio into one message.** It would change the `settings:changed` payload shape across a contract boundary and all 13 subscribers, on a production app, to save 2 small IPC messages per apply. With the identity check the trio's render cost is already zero when nothing differs, so the remaining cost is not worth a contract change. Flagging for a decision rather than silently skipping.
- **`use-calendar-range.ts`** invalidates a React Query key on any `calendar` settings change, including its own echo. It holds no comparable local state, so the same technique does not apply; left alone to keep this change surgical.


h4yfans:
Follow-up commit `1868eb18f`: `codecov/patch` flagged the changed listener line in eight of the settings hooks — `useGeneralSettings`, `useEditorSettings`, `useJournalSettings`, `useTabPreferences`, `useInboxPreferences`, `useKeyboardSettings`, `useNoteEditorSettings` and `useCalendarPreferences` had no test exercising it (only `useTaskPreferences` and `useFeatureFlags` did).

Added `src/renderer/src/hooks/settings-echo-dedupe.test.ts`, a table-driven regression that mounts each hook, echoes the unchanged group back and asserts state identity is preserved, plus that a genuinely different echo still applies and other group keys are ignored. Covering every subscriber rather than one representative is deliberate: the regression is per-hook, so a single hook reverting to `{ ...prev, ...value }` would silently re-render its consumers on every sync apply again.

Verified the changed lines are actually covered rather than merely executed — parsed the istanbul JSON report against `git diff -U0 origin/main`:

```
OK  apps/desktop/src/renderer/src/contexts/tabs/reducers/session-reducer.ts
OK  apps/desktop/src/renderer/src/hooks/use-calendar-preferences.ts
OK  apps/desktop/src/renderer/src/hooks/use-editor-settings.ts
OK  apps/desktop/src/renderer/src/hooks/use-feature-flags.ts
OK  apps/desktop/src/renderer/src/hooks/use-general-settings.ts
OK  apps/desktop/src/renderer/src/hooks/use-inbox-preferences.ts
OK  apps/desktop/src/renderer/src/hooks/use-journal-settings.ts
OK  apps/desktop/src/renderer/src/hooks/use-keyboard-settings.ts
OK  apps/desktop/src/renderer/src/hooks/use-note-editor-settings.ts
OK  apps/desktop/src/renderer/src/hooks/use-tab-preferences.ts
OK  apps/desktop/src/renderer/src/hooks/use-task-preferences.ts
OK  apps/desktop/src/renderer/src/lib/settings-patch.ts

=> all changed source lines covered
```

Mutation-tested as well: reverting `mergeSettingsPatch` to an unconditional spread fails 8 of the new file's 24 tests. Full targeted re-run is 158 passed (15 files); `typecheck:web` exit 0, eslint clean, `git diff --check` clean.

No production code changed in this commit — tests only.

PR #1226, opened 2026-08-11
fix(home): persist widget layout at every window width

### LINKED ISSUE #1216: [LOW][home] Home page widget layout is not persisted — resets on every launch, both macOS and Linux
## Where this came from

**Jerry — 2026-08-11, `v2026-08-11`:**

> Minor thing...Home page does not save my layout if I resize and move things around. Have to tweak it upon launch on both Mac and Fedora.

His words, "minor" — but note the cost: he re-arranges the Home page **on every launch**, on two machines.

## What's reported

Resizing and moving Home page widgets does not persist. The layout resets to its default on the next launch, on both macOS and Linux, on the current release.

Not yet diagnosed. Both platforms behave the same, so this is unlikely to be an OS-level storage problem — more likely the layout is either never written, written somewhere that startup doesn't read, or overwritten by a default on mount.

## What to check

1. Whether a Home layout is persisted at all on drag/resize end, and where — vault config, settings, or renderer-local storage.
2. If it is written: whether startup reads it before or after the default grid is applied.
3. Whether the widget grid remounts during launch and re-seeds defaults over a restored layout.
4. Whether the layout is expected to sync across devices. He uses two machines and mentions both, so it is worth deciding whether this is per-device state or synced state before fixing the persistence.

## Why it matters

Small, repeated friction on the first screen the user sees every session. The Home dashboard is a customisation surface whose whole point is that it stays the way you left it, so the failure lands every single launch.

h4yfans:
Closes #1216

## Root cause

The layout **was** being written (data DB `home_pages.widgets`, via `homePages.update` IPC) — but only for edits made at one window width.

`board-grid.tsx` rendered react-grid-layout's **Responsive** grid with `breakpoints = {lg: 1024, md: 768, sm: 0}` and `cols = {lg: 8, md: 4, sm: 2}`. A board stores a *single* arrangement (`x/y/w/h` per widget), so the handler deliberately persisted only `allLayouts.lg` — persisting a collapsed narrow layout would have overwritten the real arrangement.

RGL only ever writes the **active** breakpoint into `allLayouts` (`dist/chunk-WGL5FSZH.mjs`: `handleLayoutChange` → `{...currentLayouts, [breakpoint]: newLayout}`). So below the `lg` breakpoint, `allLayouts.lg` is byte-identical to what we passed in, the `unchanged()` guard short-circuits, and **every drag and resize is dropped on the floor**. The move looks fine for the session (RGL's own state) and is gone on the next launch.

The grid is the window minus the sidebar (256px) and page padding (48px), so `lg` needs a window wider than roughly **1330px**. Any normal non-maximized window — on macOS and on Fedora — sits at `md`, which is exactly what Jerry reports on both machines.

## Fix

One fixed **8-column** grid at every width (`GRID_COLS = 8`, legacy `GridLayout` instead of `Responsive`). Columns scale with the container instead of collapsing, so what RGL reports *is* the board's arrangement and can always be persisted. A widget keeps the same share of the board on a 13" laptop and a 27" display, which is a better fit for a two-machine user than a reflow that only existed to be thrown away.

Backward compatible: stored coordinates are unchanged (same 8-column space), no migration, no contract or schema change. Existing boards render exactly as authored; on a narrow window they now scale down instead of reflowing to 4/2 columns.

## Per-device or synced?

Today Home boards are **per-device**: `home_page` is not in `SYNC_ITEM_TYPES`, there is no item handler, nothing is pushed. Jerry sees the bug on two machines because each machine drops its own edits independently — not because the two disagree.

**They should sync.** A board is authored content (which widgets, their configs — a Tasks widget points at a saved filter, and filters already sync), not device state; and now that positions are relative to a fixed 8-column grid rather than pixel-bound, an arrangement transfers cleanly between differently sized screens. That is a sync-protocol addition (new item type + handler + field-level clocks, following `folder_config`) and does **not** belong inside a persistence bugfix — filing it as a follow-up.

The docs claimed boards sync; corrected to say they are per-device for now.

## Verification

- New `board-grid-persistence.test.tsx` drives the **real** grid — no RGL mock, real `mousedown`/`mousemove`/`mouseup` drag — at a deliberately narrow board width, then unmounts and remounts from what was persisted. Fails on `main` (`x: 0` — the drag moved the widget on screen but nothing was saved), passes here.
  - The pre-existing `board-grid.test.tsx` mocked RGL entirely, so it asserted the lg-only rule was implemented correctly rather than that a user's drag survives. Its obsolete breakpoint case is replaced.
  - Two jsdom stubs, both narrow and documented: `WidthProvider` (its only job is measuring, which jsdom reports as 0) and `HTMLElement.offsetParent` (null in jsdom, and RGL refuses to start a drag without it).
- `pnpm lint` (0 errors), `pnpm typecheck`, home renderer suite (18 files / 74 tests), `pnpm docs:impact --base origin/main --strict`, `pnpm docs:build`.

## Not verified

See the verification comment below: the branch has since been rebased onto current `main` and exercised in a running Electron app, including an A/B of the narrow-window rendering against the pre-fix build.

### DIFF
--- a/apps/desktop/src/renderer/src/components/home/board-grid-persistence.test.tsx
+++ b/apps/desktop/src/renderer/src/components/home/board-grid-persistence.test.tsx
@@ -1,0 +1,115 @@
+import { useState } from 'react'
+import { describe, it, expect, vi, beforeAll } from 'vitest'
+import { render, screen, fireEvent } from '@testing-library/react'
+import { BoardGrid } from './board-grid'
+import { TooltipProvider } from '@/components/ui/tooltip'
+import { registerWidget } from '@/lib/home/widget-registry'
+import { GRID_MARGIN, GRID_ROW_HEIGHT } from '@/lib/home/widget-sizes'
+import type { HomePage } from '@/lib/home/types'
+
+// Board width the grid believes it has. Deliberately NARROW (a 1280px window minus the sidebar and
+// page padding lands around here) — the regression this file guards is that a drag made at a narrow
+// window width was silently dropped. Only WidthProvider is stubbed, and only because its single job
+// is measuring the container, which jsdom always reports as 0. The grid itself is the real one.
+const BOARD_WIDTH = 900
+
+vi.mock('react-grid-layout/legacy', async (importOriginal) => {
+  const actual = await importOriginal<typeof import('react-grid-layout/legacy')>()
+  return {
+    ...actual,
+    WidthProvider:
+      <P,>(Component: React.ComponentType<P & { width: number }>) =>
+      (props: P) => <Component {...props} width={BOARD_WIDTH} />
+  }
+})
+
+// jsdom never lays anything out, so `offsetParent` is null and react-grid-layout refuses to begin a
+// drag. Point it at the parent element so the drag pipeline runs for real.
+beforeAll(() => {
+  Object.defineProperty(HTMLElement.prototype, 'offsetParent', {
+    configurable: true,
+    get(this: HTMLElement) {
+      return this.parentElement
+    }
+  })
+})
+
+registerWidget({
+  type: 'bookmarks',
+  titleKey: 'Bookmarks',
+  icon: 'bookmark',
+  defaultLayout: { w: 2, h: 2 },
+  defaultConfig: {},
+  Component: () => <div>BM</div>
+})
+
+const initialBoard: HomePage = {
+  id: 'b1',
+  name: 'B',
+  position: 0,
+  widgets: [{ id: 'w1', type: 'bookmarks', x: 0, y: 0, w: 2, h: 2, config: {} }]
+}
+
+// Stands in for the persisted board (home_pages row). BoardGrid writes through onChange; a remount
+// reads back whatever was written, so the assertions cover a real save→reload round trip.
+function makeStore(board: HomePage) {
+  return { board }
+}
+
+function Harness({ store }: { store: { board: HomePage } }): React.JSX.Element {
+  const [board, setBoard] = useState(store.board)
+  return (
+    <TooltipProvider>
+      <BoardGrid
+        board={board}
+        onChange={(next) => {
+          store.board = next
+          setBoard(next)
+        }}
+      />
+    </TooltipProvider>
+  )
+}
+
+function widgetPosition(): { x: number; y: number } {
+  const el = document.querySelector<HTMLElement>('.react-grid-item')
+  const transform = el?.style.transform ?? ''
+  const match = /translate\((-?[\d.]+)px,\s*(-?[\d.]+)px\)/.exec(transform)
+  if (!match) throw new Error(`no translate on grid item: ${transform}`)
+  return { x: Number(match[1]), y: Number(match[2]) }
+}
+
+// One grid column / row step in pixels, so drags move a whole cell.
+const colStep = (BOARD_WIDTH - GRID_MARGIN[0] * 9) / 8 + GRID_MARGIN[0]
+const rowStep = GRID_ROW_HEIGHT + GRID_MARGIN[1]
+
+function dragWidget(dx: number, dy: number): void {
+  const handle = screen.getByLabelText('Drag widget')
+  fireEvent.mouseDown(handle, { clientX: 0, clientY: 0, button: 0 })
+  fireEvent.mouseMove(document, { clientX: dx, clientY: dy })
+  fireEvent.mouseUp(document, { clientX: dx, clientY: dy })
+}
+
+describe('BoardGrid persistence', () => {
+  it('persists a drag and restores it on remount', () => {
+    const store = makeStore(initialBoard)
+    const first = render(<Harness store={store} />)
+
+    const before = widgetPosition()
+    dragWidget(Math.round(colStep * 3), Math.round(rowStep * 2))
+    // Sanity: the drag really moved the widget on screen (guards against a no-op simulation).
+    expect(widgetPosition().x).toBeGreaterThan(before.x)
+
+    // The move reached the persisted board, not just react-grid-layout's internal state.
+    expect(store.board.widgets[0]).toMatchObject({ id: 'w1', x: 3 })
+    const moved = store.board.widgets[0]
+
+    first.unmount()
+
+    // Remount from what was persisted — the arrangement must come back, not reset to the default.
+    render(<Harness store={makeStore(store.board) as { board: HomePage }} />)
+    expect(store.board.widgets[0]).toMatchObject({ x: moved.x, y: moved.y })
+    const pos = widgetPosition()
+    expect(pos.x).toBeGreaterThan(0)
+  })
+})
--- a/apps/desktop/src/renderer/src/components/home/board-grid.test.tsx
+++ b/apps/desktop/src/renderer/src/components/home/board-grid.test.tsx
@@ -9,16 +9,16 @@ import type { HomePage } from '@/lib/home/types'
 // react-grid-layout measures container width via ResizeObserver (0 in jsdom), so stub it to a
 // passthrough that renders children and captures onLayoutChange so tests can fire RGL callbacks.
 const rgl = vi.hoisted(() => ({
-  onLayoutChange: undefined as undefined | ((layout: unknown, all: unknown) => void)
+  onLayoutChange: undefined as undefined | ((layout: unknown) => void)
 }))
 vi.mock('react-grid-layout/legacy', () => ({
   WidthProvider: (C: unknown) => C,
-  Responsive: ({
+  default: ({
     children,
     onLayoutChange
   }: {
     children: React.ReactNode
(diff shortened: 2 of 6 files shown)


h4yfans:
## Rebased + verified in a running app

Rebased onto `origin/main` (`7436b3321`) — the branch was 126 commits behind, so the earlier green CI was against a stale base. **No conflicts**; the diff is unchanged (5 files, +144/−40).

### The narrow-window trade-off — now measured, not assumed

This PR deliberately drops the responsive grid, so I ran the real app (Playwright + Electron, fresh vault, default seeded board plus a Tasks and a Calendar widget) at eight window widths, and then rebuilt with the pre-fix source and ran the identical script, so the two columns below are the same app at the same widths with only `board-grid.tsx` + `widget-sizes.ts` differing.

| window width | board width | widget on `main` | widget on this branch |
| --- | --- | --- | --- |
| 1550px (default) | 915px | 891x260 | 440x260 |
| 1328px | 693px | 669x260 | 329x260 |
| 1100px | 465px | 441x260 | 215x260 |
| 1000px | 365px | 341x260 | 165x260 |
| 900px | 265px | 241x260 | 115x260 |
| 800px | 165px | 141x260 | 65x260 |
| 700px | 321px | 297x260 | 143x260 |
| 600px | 221px | 197x260 | 93x260 |

Two things this changes about the writeup above.

**1. The bug is worse than the PR body says — it hits the default window.** With the sidebar *and* the Day Panel open (both default), the board is only **915px at the default 1550px window**. That is already below `GRID_BREAKPOINTS.lg = 1024`, so on `main` *every* user is sitting at `md`/`sm` all the time, not just narrow-window users. That matches the `main` column: a `w:4` widget renders full-board-width at every size tested, i.e. the stored 8-column arrangement is never what you see and never what gets saved. (The 700px row is non-monotonic because the sidebar auto-collapses between 800 and 700 and hands ~256px back to the board.)

**2. There is a real visual cost below roughly a 1000px window.** Row height is fixed (`GRID_ROW_HEIGHT = 56`), so as columns scale down, widgets get thin without getting shorter:

- **1550px / 1328px** — good. Widgets sit at their true half-board width, which is the arrangement the user actually authored.
- **1100px / 1000px** — cramped but usable; widget titles start truncating (`RECENTL…`, `BOOKMA…`).
- **900px** — poor; titles are down to one letter (`R…`, `B…`) and the Tasks filter chip clips.
- **800px** (sidebar still expanded, Day Panel open) — **worst case: 65x260**. The title is gone entirely, only the icon remains, "No notes yet." wraps one word per line, and the content area gains a horizontal scrollbar.

So: better at the sizes people actually use, worse below ~1000px, and the ~800px case is genuinely bad. It is a real regression at those widths — on `main` a widget stays board-width there — but `main` buys that legibility by discarding the arrangement, which is the bug being fixed.

Nothing floors how far this can go: there is no `minWidth` on the `BrowserWindow` (`apps/desktop/src/main/index.ts:666`), so the window can be dragged to any width.

**Recommended mitigation — not implemented here, it is your call.** The smallest contained fix is a width floor on the board rather than on the window: a `min-width` of ~640px on `.home-grid` (`apps/desktop/src/renderer/src/components/home/home-grid.css`). At 640px a column is 68px, so a `w:4` widget never drops below ~296px — about the 700px-window quality — and the page container (`home.tsx`, already `overflow-auto`) scrolls horizontally below that. It changes nothing about persistence: RGL still reports the same 8-column space. I did not apply it because trading reflow for horizontal scrolling is a design decision, and a `BrowserWindow.minWidth` alternative would have to be ~1265px to achieve the same thing, which is too aggressive for small displays.

### Verification (all re-run after the rebase)

| check | result |
| --- | --- |
| `pnpm lint` | exit 0 — 0 errors, 28 pre-existing warnings, none in the touched files |
| `pnpm typecheck` | exit 0 |
| `pnpm --filter @memry/desktop build` | exit 0 (full desktop typecheck + bundle + worker-bundle check) |
| Home renderer suite (`--project renderer src/renderer/src/components/home src/renderer/src/lib/home`) | 17 files / 75 tests passed |
| `pnpm docs:impact --base origin/main --strict` | exit 0 — `covered against origin/main` |
| `pnpm docs:build` | exit 0 |

**Mutation-checked the regression test.** With `board-grid.tsx` and `widget-sizes.ts` reverted to `origin/main` and the new test file left in place, `board-grid-persistence.test.tsx` fails exactly as it should:

```
AssertionError: expected { id: 'w1', type: 'bookmarks', …(5) } to match object { id: 'w1', x: 3 }
```

The drag moves the widget on screen and nothing reaches the persisted board. It is a genuine regression test, not one that passes either way.

### Still unverified

- Only exercised on macOS. Not run on Linux or Windows, so the reporter's Fedora machine is still inferred from the shared code path rather than observed.
- The E2E harness resizes the window via `BrowserWindow.setSize`; a user dragging the window edge produces a continuous stream of resizes rather than discrete steps. Not tested.
- Home boards are still per-device (`home_page` is not in `SYNC_ITEM_TYPES` and has no item handler) — confirmed against the rebased tree, so the docs correction in this PR is accurate. Syncing them remains a follow-up.

The screenshots behind the table are on this machine at `/tmp/memry-narrow-board/board-<width>.png` (this branch) and `/tmp/memry-narrow-board-main/board-<width>.png` (pre-fix), with the throwaway capture script saved next to them as `/tmp/memry-narrow-board/narrow-board-visual.e2e.ts`. It was not committed — it only screenshots and measures, it asserts nothing.

h4yfans:
## Width floor added (`64645d457`)

The single change, in `apps/desktop/src/renderer/src/components/home/home-grid.css`:

```css
.home-grid {
  min-width: 640px;
}
```

Nothing else touched: `BrowserWindow.minWidth`, `GRID_COLS`, `MIN_W`, `GRID_ROW_HEIGHT` and the persistence path are all untouched. RGL derives its column width from this element's measured width, so the floor is simply what it computes against — stored 8-column coordinates are unchanged and the fix in the first commit still behaves identically.

The user-guide sentence about columns narrowing was extended to mention that they stop narrowing at the minimum and the board scrolls sideways from there.

### Measured, same script, same widths

Board width now clamps at 640px and a four-column widget holds at **302px** everywhere below the floor. Above the floor nothing changes — 1550px and 1328px are byte-identical to the no-floor run.

| window | board (with floor) | `main` | branch, no floor | branch + floor |
| --- | --- | --- | --- | --- |
| 1550px | 915px | 891x260 | 440x260 | **440x260** |
| 1328px | 693px | 669x260 | 329x260 | **329x260** |
| 1100px | 640px | 441x260 | 215x260 | **302x260** |
| 1000px | 640px | 341x260 | 165x260 | **302x260** |
| 900px | 640px | 241x260 | 115x260 | **302x260** |
| 800px | 640px | 141x260 | 65x260 | **302x260** |
| 700px | 640px | 297x260 | 143x260 | **302x260** |
| 600px | 640px | 197x260 | 93x260 | **302x260** |

**800px** — the case that was 65px wide with no title at all: titles now render in full (`RECENTLY EDITED`, `BOOKMARKS`, `TASKS` with its `Due Today` chip) and body text sits on one line instead of wrapping one word per line. The board extends past the viewport and you scroll to reach the rest.

**1000px** — full titles, the widget menu and the task count both visible, `No notes yet.` on a single line. Previously this truncated to `RECENTL…` at 165px wide.

Worth noting the floor beats `main` here too, not just the no-floor branch: at 800px `main` also degraded to `RECE…`/`BOO…` because its collapsed 2-column widget was only 141px.

### One correction to the horizontal-scroll assumption

The premise holds at 800px — I checked `main`'s own 800px capture and it already shows a horizontal scrollbar, so the floor is not what introduces scrolling there.

It does **not** hold across the whole range, and I would rather flag it than let it surprise you. The floor engages as soon as the natural board width drops under 640px, which happens at roughly a **1270px window** with the sidebar and Day Panel open. Between there and ~800px the floor introduces a horizontal scrollbar where neither `main` nor the no-floor branch had one — at a 1000px window the board would naturally be 365px, and it is now a 640px grid inside a 365px container. That is the intended trade (scroll rather than illegible widgets) and it is what makes the 302px number possible, but it is new behaviour in that band rather than pre-existing. The code comment says so explicitly.

### Gate

| check | result |
| --- | --- |
| `pnpm lint` | exit 0 — 0 errors, 28 pre-existing warnings, none in the touched files |
| `pnpm typecheck` | exit 0 — 16/16 tasks (full turbo cache hit; this commit changes only CSS and Markdown, neither a TypeScript input) |
| Home renderer suite (`--config config/vitest.config.ts --project renderer …`) | 17 files / 75 tests passed |
| `pnpm docs:impact --base origin/main --strict` | exit 0 — `covered against origin/main`, `home-grid.css` picked up as docs-relevant |
| `pnpm docs:build` | exit 0 |

Full `pnpm test` was not run. The capture script remains uncommitted; new screenshots are at `/tmp/memry-narrow-board-floor/board-<width>.png`.

Still only verified on macOS.

Issue #993, opened 2026-08-07
[MAJOR][projections] Full-vault FTS teardown + disk re-scan on every vault open; not abortable, keeps reading the old vault after a switch


h4yfans:
Part of epic #987 — Memory & CPU full-app performance audit 2026-08.

> [!IMPORTANT]
> **For the implementing agent: validate before implementing.**
> This issue was filed from a static audit on 2026-08-07 (`main` @ e4c9c90ee). Line numbers may have drifted.
> 1. First, CONFIRM the issue is still valid on current `main`: read the referenced files, verify the described behavior still exists in code, and check that no later commit already fixed it.
> 2. If it is no longer valid, comment your evidence and close the issue — do not implement.
> 3. If valid, implement following the conventions in `CLAUDE.md`/`AGENTS.md`: surgical change, tests included, run the Verify commands (`pnpm lint`, `pnpm typecheck`, relevant `pnpm test:*`, `pnpm docs:impact`) before opening a PR that references this issue.
> 4. Where a memory/CPU delta is measurable, attach before/after evidence (see #446 memory snapshot harness).

---

**Problem.** `apps/desktop/src/main/projections/projectors/search-projector.ts:255-259` — `reconcile()` calls `rebuildNotes` (`:82-136`: `DELETE FROM fts_notes`, then `fs.readFile` + `parseNote` + `insertFtsNote` for every markdown row), plus `rebuildTasks()` and `rebuildInbox()`. Invoked via `void reconcileProjections()` on **every** `openVault` (`apps/desktop/src/main/vault/index.ts:407`).

**Why it matters.** On a vault with thousands of notes this is thousands of file reads, full markdown parses, and a complete FTS5 rewrite on every launch — for an index that is already correct. It is fire-and-forget with no abort: `stopProjectionRuntime` only nulls the module reference (`projections/index.ts:47-49`), so switching vaults mid-reconcile keeps reading the **old** vault path against a now-closed database, logging one warning per note (`:122-124`).

**Trigger.** Every app launch and vault switch.

**Suggested fix.** Make reconcile a cheap consistency check (compare mtime/hash or row counts; rebuild only divergent entries). Thread an abort token through reconcile and cancel it in `stopProjectionRuntime`.

**Verification.** Launch with a warm index performs no full FTS rewrite; switching vaults mid-reconcile stops the old pass (no old-vault reads after switch).

Issue #805, opened 2026-07-20
[Bug] Windows update refuses to install ("MemryNote cannot be closed") — embeddings utility process not terminated on quit


h4yfans:
**Reported by:** Aurelie Kabore — Windows 10, v2026.718.3
**Area:** Auto-update / NSIS installer / process lifecycle
**Verdict:** Partially real (install-refusal mechanism confirmed in code; "no auto-launch after install" is expected, not a bug)

## Symptom
"The update was refusing to install with same message as to close Memrynote though I hadn't opened... nor was there in process to end it." She had to uninstall to update.

## Root cause
The dialog is electron-builder's NSIS `appCannotBeClosed` string, raised after `CHECK_APP_RUNNING` finds a live `Memrynote.exe` and can't close it within ~5x1s retries (detection by exe name `Memrynote.exe`, `config/electron-builder.yml:39`). A phantom `Memrynote.exe` can linger because:
1. **Primary:** the embeddings `utilityProcess` is never terminated in the `before-quit` shutdown chain. `before-quit` (`main/index.ts:1631-1748`) stops the capture server, voice model (`:1688`), image processing (`:1692`), sync — but never calls `stopEmbeddingModel()` (defined/exported at `lib/embeddings.ts:602`, with no lifecycle caller). The worker relies solely on a 30s idle self-exit (`IDLE_SHUTDOWN_MS`, `embeddings.ts:61`). On Windows `utilityProcess.fork` runs as the app binary -> the worker **is a `Memrynote.exe`**. If it's mid-rebuild at "Restart to install," it outlives the NSIS retry window.
2. `quitAndInstall()` -> electron-updater `doInstall` spawns the installer and returns immediately (`--updated /S --force-run`, `updater.ts:303,315-316`); the running-app check can run before children are reaped.
3. No Windows install guard: `updater-install-guard.ts` is entirely darwin-gated (`index.ts:942` never fires on Windows).
4. Residual invisible-zombie on the first launch after a GPU crash (guard only self-heals on the next launch).

**Not a defect:** "app didn't auto-launch after install" — assisted installer (`oneClick:false`, `allowToChangeInstallationDirectory:true`, no `runAfterFinish`, `electron-builder.yml:44-53`) ends on a finish page with a "Run" checkbox after a manual reinstall.

## Fix direction
[...]

Issue #803, opened 2026-07-20
[Bug] Opening a vault hangs indefinitely (>5 min) — embeddings run inline before the vault is marked open


h4yfans:
**Reported by:** Aurelie Kabore — Windows 10, v2026.718.3
**Area:** Vault open / indexing / embeddings
**Verdict:** Real (the customer's `.github` hypothesis is wrong — dotfolders are already excluded)

## Symptom
"I decided to open one of my Obsidian vault... I waited and waited... more than 5 minutes. I had to close and decided to try a smaller vault... Same thing, it kept loading only." (She uses the Obsidian Git plugin, so the vault has `.git`/`.github`.)

## Not the cause: .git/.github
Dotfolders are excluded from both scan and watcher: `findVaultFiles` skips `entry.name.startsWith('.')` (`main/vault/indexer.ts:74-75`); the chokidar watcher ignores `basename.startsWith('.')` (`main/vault/watcher.ts:198`), and `followSymlinks:false` rules out symlink loops.

## Root cause
`openVault` awaits `indexVault()` **before** flipping the vault open (`main/vault/index.ts:310` `await indexVault()` -> `:345` `isOpen:true`). The comment at `:339-344` claims embedding is deferred, but that only covers the backgrounded `reconcileProjections()` — the per-note `project()` path is a separate, fully-awaited blocker:
1. Each new markdown file -> `syncNoteToCache` publishes `note.upserted` (`vault/note-sync.ts:236,332`).
2. `indexer.ts:196` `await flushProjectionEvents()` drains the bus synchronously (`projections/runtime.ts:71,108-109`).
3. The embedding projector handles `note.upserted` -> `updateEmbedding()` (`embedding-projector.ts:59-79,163-188`): `initEmbeddingModel()` downloads a **~23MB model on first use** via a `utilityProcess` worker whose per-request timeout is **`REQUEST_TIMEOUT_MS = 5*60_000` = 5 minutes** (`lib/embeddings.ts:58`), then CPU inference **per note**.
4. AI is **on by default** (`isAIEnabled()` returns true when unset, `embedding-projector.ts:34-42`).

[...]

PR #1243, opened 2026-08-11
fix(projections): abort the embedding backfill on stop and drain deferred notes after a reindex

### LINKED ISSUE #1083: [MEDIUM][projections] Embedding-projector state hygiene: pendingEmbedding only drained on vault open; module-global vault path provider never reset

> [!IMPORTANT]

---

**Problem.** (1) `projections/projectors/embedding-projector.ts:86,208-210` — every note upserted while `isIndexing()` lands in `pendingEmbedding`; but `updateConfig` structural changes (`vault/index.ts:552-562`) and `reindex()` (`:673-688`) set `isIndexing` and run a full pass **without** calling `reconcileProjections()` after — only `openVault` does (`:407`). The set retains one id per touched note until the next vault open. (2) `note-derived-state-projector.ts:128,133` — `currentVaultPathProvider` is set as a construction side effect and never reset in `stopProjectionRuntime`/`closeVault`; retains the last vault path closure and makes the projector non-reentrant.

**Suggested fix.** Drain pending after any full pass completes; reset the module global on runtime stop (or eliminate it — pass the provider through the projector instance).

**Verification.** Reindex → pending set empty; vault close nulls the provider.

h4yfans:
Part of epic #987 (memory & CPU audit 2026-08). Closes #1083.

## Validation first

The issue described two defects. One is already fixed on `main`, one is still live — and the live one turned out to have a worse shape than the audit described.

**(2) module-global vault path provider — ALREADY FIXED, not re-implemented.** At audit time (`e4c9c90ee`) `note-derived-state-projector.ts` held `let currentVaultPathProvider` at `:128`, written as a construction side effect at `:133` and read at `:114`. Commit `f34d4cb04` ("refactor: add projection pipeline runtime") eliminated it: the provider is now a per-instance closure parameter (`createNoteDerivedStateProjector(getVaultPath)`, `note-derived-state-projector.ts:213`), and `grep -rn currentVaultPathProvider apps/desktop/src` returns nothing. No change needed.

**(1) `pendingEmbedding` only drained on vault open — STILL VALID.** Confirmed below, and fixed.

## What was wrong

**Primary: the embedding projector ignored its `AbortSignal`.**

`ProjectionProjector.reconcile(signal?: AbortSignal)` (`projections/types.ts:91`) exists precisely so a stopping runtime can cut a backgrounded pass short (#993). `search-projector.ts` honours it at ~10 checkpoints. `embedding-projector.ts:221` declared `async reconcile(): Promise<void>` — it never accepted the signal at all. It is also the most expensive projector: a ~23MB model load plus CPU inference per note.

Two consequences:

- **Vault close/quit was held open.** `closeVault()` awaits `stopProjectionRuntime({drain:true})` (`vault/index.ts:603`); `runtime.stop()` aborts and then `await pendingReconcile` (`runtime.ts:281-283`) before the caller closes the databases. With the abort ignored, that await ran the *entire* remaining work list — model load included. This is exactly the stall class of #803 / #805.
- **Cross-vault writes on a vault switch.** `startProjectionRuntime` supersedes a live runtime with `void supersededRuntime.stop({ drain: false })` — deliberately not awaited (`projections/index.ts:38`). Its comment reasons carefully about one in-flight `project()` event leaking a row, on the assumption the reconcile abort works. It did not. The old pass kept reading vault A's files via its own `getVaultPath` closure and writing them through the *global* `getRawIndexDatabase()`, which by then points at vault B's index. Same consequence the audit attributed to the module global, arriving by a different route.

**Secondary (the filed issue): deferred ids never drained without a restart.** Every note upserted while `isIndexing()` lands in `pendingEmbedding` (`embedding-projector.ts:208-210`) instead of embedding inline. Only `openVault` reconciled afterwards (`vault/index.ts:430`). `updateConfig`'s structural rebuild (`:575-585`) and `reindex()` (`:739-754`) both set `isIndexing`, run a full pass, clear it — and stop. Those notes kept a stale or missing vector, and their ids sat in the set, until the next vault open.

## What changed

`embedding-projector.ts` — `reconcile(signal)`, `runRebuild(signal)` and `embedNotes(…, signal)` check the signal at every await boundary (per note in the embed loop, after the model load, around the version-migration rebuild). An aborted pass:
- keeps its deferred ids — the work is owed, not lost, and the next pass picks it up;
- does not run `DELETE FROM vec_notes` it is no longer going to refill;
- does not stamp `ai.embeddingInputVersion` for a partial migration rebuild (a partial stamp would permanently skip the migration on an existing install).

`vault/index.ts` — new `drainDeferredEmbeddings()` called after the `reindex()` pass and after `updateConfig`'s structural rebuild.

## Drain policy, and why

**Backgrounded `reconcileProjections(['embedding'])`, fired after the pass completes — not a timer, not vault close.**

- *Not at vault close:* close is the one path that must never grow work. Draining there re-creates #803/#805 by design.
- *Not on a timer/idle:* `ProjectionProjector` has no dispose hook, so a projector-owned timer would have to be torn down through a new lifecycle seam, and a stray one keeps the event loop alive at quit. The drain is not periodic work — it has exactly two missing trigger points.
- *At the end of the pass* is where `openVault` already drains (`vault/index.ts:430`), so this reuses a proven pattern with the same shape: `void … .catch()`, never awaited, so neither `reindex()` nor `updateConfig()` waits on a model load.
- *Scoped to `['embedding']`* rather than a full reconcile: this is a CPU-audit fix, and running every projector's full consistency pass on each config save would be a regression in spirit. Only the embedding projector's state is at issue.

The two halves interlock: the drain is only safe to leave running in the background *because* the abort fix makes it stop on demand.

## How I proved vault close is not blocked

`lets the projection runtime stop without waiting for a large backfill` drives the real `createProjectionRuntime` with the real embedding projector over a 200-note work list, starts `runtime.reconcile()`, then awaits `runtime.stop({ drain: false })` — the exact call `closeVault` makes — and asserts fewer than 200 notes were embedded. No timers, no sleeps: without the fix `stop()` resolves only after all 200, so the assertion fails deterministically.

## Tests

5 new tests, all mutation-checked: with the two source files reverted to `HEAD` and only the tests applied, exactly these 5 fail (`Tests  5 failed | 22 passed`); with the fix, `Tests  27 passed`.

- abort mid-work-list stops the pass, leaves the unreached notes' files unread (so nothing can land in the next vault's index), and a later reconcile embeds them
- an already-aborted signal loads no model and touches no database
- an abort during the version-migration rebuild neither wipes `vec_notes` nor stamps the version
- the runtime stop test above
- `vault/index.test.ts`: `reindex()` and a structural `updateConfig()` each drain via `reconcileProjections(['embedding'])`

## Verification

```
pnpm --filter @memry/desktop typecheck:node     # clean
pnpm exec eslint <4 changed files>              # 0 errors
vitest --project main src/main/projections      # 8 files passed
vitest --project main …embedding-projector.test.ts …vault/index.test.ts
                                                # Tests  27 passed (27)
git diff --check                                # clean
pnpm docs:impact --base origin/main --strict    # covered
pnpm docs:build                                 # build complete
```

Backward compatibility: no schema, contract, or format change. Behaviour on an existing install is strictly additive — passes that used to run to completion still do; they can now also stop early, and early stops preserve rather than discard pending work.

### DIFF
--- a/apps/desktop/src/main/projections/projectors/embedding-projector.test.ts
+++ b/apps/desktop/src/main/projections/projectors/embedding-projector.test.ts
@@ -401,6 +401,67 @@ describe('embedding projector', () => {
     expect(generateEmbedding).not.toHaveBeenCalled()
   })
 
+  it('does not stamp the embedding input version when the migration rebuild aborts mid-list', async () => {
+    const run = vi.fn()
+    const prepare = vi.fn(() => ({ run, all: () => [] as Array<{ note_id: string }> }))
+    getRawIndexDatabase.mockReturnValue({ prepare })
+    getIndexDatabase.mockReturnValue({
+      all: vi.fn(() => [
+        { id: 'note-1', path: 'notes/one.md', title: 'One', fileType: 'markdown' },
+        { id: 'note-2', path: 'notes/two.md', title: 'Two', fileType: 'markdown' }
+      ])
+    })
+    getSetting.mockImplementation((_db: unknown, key: string) =>
+      key === 'ai.embeddingInputVersion' ? 'stale' : 'true'
+    )
+    readFile.mockResolvedValue('raw markdown')
+    parseNote.mockReturnValue({ content: 'parsed markdown long enough' })
+
+    const controller = new AbortController()
+    generateEmbedding.mockImplementation(async () => {
+      controller.abort()
+      return new Float32Array([0.1, 0.2])
+    })
+
+    const projector = createEmbeddingProjector(() => '/vault')
+
+    await projector.reconcile(controller.signal)
+
+    // A half-finished rebuild must not record the new input version: doing so
+    // would permanently skip the migration for the notes it never reached.
+    expect(generateEmbedding).toHaveBeenCalledTimes(1)
+    expect(setSetting).not.toHaveBeenCalled()
+  })
+
+  it('skips the backfill when the signal aborts while the model is loading', async () => {
+    const run = vi.fn()
+    const prepare = vi.fn(() => ({ run, all: () => [] as Array<{ note_id: string }> }))
+    getRawIndexDatabase.mockReturnValue({ prepare })
+    getIndexDatabase.mockReturnValue({
+      all: vi.fn(() => [{ id: 'note-1', path: 'notes/one.md', title: 'One' }])
+    })
+    getSetting.mockImplementation((_db: unknown, key: string) =>
+      key === 'ai.embeddingInputVersion' ? String(EMBEDDING_INPUT_VERSION) : 'true'
+    )
+
+    // The load is the multi-second step, so a close landing inside it is the
+    // common case — the work list must not be embedded once it returns.
+    const controller = new AbortController()
+    isModelLoaded.mockReturnValue(false)
+    initEmbeddingModel.mockImplementation(async () => {
+      controller.abort()
+      return true
+    })
+
+    const projector = createEmbeddingProjector(() => '/vault')
+
+    await projector.reconcile(controller.signal)
+
+    expect(initEmbeddingModel).toHaveBeenCalled()
+    expect(readFile).not.toHaveBeenCalled()
+    expect(generateEmbedding).not.toHaveBeenCalled()
+  })
+
   it('lets the projection runtime stop without waiting for a large backfill', async () => {
     const noteCount = 200
     const run = vi.fn()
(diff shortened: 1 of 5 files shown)


h4yfans:
Follow-up commit for `codecov/patch` (was 81.48%, 5 missing lines — now passing):

- dropped a redundant `signal?.aborted` guard at the top of `runRebuild` — `reconcile` already returns before it with no await in between, so the line was unreachable rather than untested;
- 3 more tests covering the remaining branches: an abort landing *mid work list* during the version-migration rebuild (still must not stamp `ai.embeddingInputVersion`), an abort landing *inside the model load* on the backfill path, and a failed background drain surfacing through `trackMainError` instead of rejecting the user's reindex.

8 new tests total, all still mutation-checked. Full run: `Tests 30 passed (30)`.

PR #1256, opened 2026-08-12
fix(tooling): stop the secret scanner reading fts5 tokenize= as a credential

### LINKED ISSUE #1143: [MEDIUM][tooling] Secret scanner reads fts5 `tokenize=` as a leaked credential and reddens CI
## Where this came from

Epic #987 → **MAJOR tier** (#988–#1019), implemented 2026-08-07/08 as PRs #1109–#1140.

This surfaced while fixing #993 (*Full-vault FTS teardown + disk re-scan on every vault open*) in **PR #1114**. That PR touches the three FTS modules, and pushing it turned the repo's own `Secret scan` **CI job red** — not just the local pre-commit hook. The workaround shipped in PR #1114 was to reformat the SQL so the offending token no longer starts a line; the scanner itself was left alone deliberately, since it is shared tooling.

Agent memory slug: `memry-ci-real-gates-and-secret-scanner-wart`

## The defect

`scripts/check-staged-secrets.mjs` flags SQL DDL as a leaked credential.

`secretAssignmentPattern` matches any **line-initial key containing `TOKEN`**, case-insensitively. The word `tokenize` contains it. So this, in an fts5 virtual-table definition:

```sql
tokenize='porter unicode61'
```

is read as a high-risk secret assignment. The usual escape hatch does not apply either: `isSourceCodeReferenceValue` bails on quoted values, so `'porter unicode61'` is never exempted.

## Why it is worth fixing rather than living with

- It fires for **anyone** who stages a file containing such a line. Today that is the three FTS DDL blocks in `apps/desktop/src/main/database/fts.ts`, `fts-tasks.ts`, `fts-inbox.ts`.
- It is not a local-only annoyance. `security-ci.yml` runs the identical script (`--changed "$BASE_REF"`), so the PR check goes red too and `--no-verify` does not get you past it.
- The failure mode teaches the wrong habit: the fastest way past it is `git commit --no-verify`, i.e. bypassing the secret scanner entirely. A scanner that people routinely bypass stops being a control.
- The current fix is cosmetic whitespace in SQL. A reformat, a prettier rule change, or a new fts5 table re-breaks CI with no obvious cause.

## What to do

Tighten the pattern so an identifier that merely *contains* `token` as a substring is not treated as a secret assignment — require a word boundary, or exclude known-safe identifiers (`tokenize`, `tokenizer`, `tokenized`, `tokenCount`, …), or exempt values that are obvious literal configuration rather than credential-shaped.

Add a regression test with a real fts5 `CREATE VIRTUAL TABLE` block asserting the scanner exits 0, and keep a test asserting an actual `API_TOKEN=...` assignment is still caught — the point is to narrow it, not blunt it.

While there, consider whether the "reads whole staged files" behaviour is right; the same property makes the renderer guard flag unrelated pre-existing lines in a staged file.

h4yfans:
Follow-up from epic #987 (MAJOR tier). Surfaced in PR #1114 (issue #993).

## The false positive

`scripts/check-staged-secrets.mjs` matched any line-initial key that merely *contained* `SECRET` / `TOKEN` / `PASSWORD` / ... as a substring, so an fts5 table option read as a leaked credential:

```
apps/desktop/src/main/database/fts.ts:31 high-risk-secret-assignment - tokenize assignment
apps/desktop/src/main/database/fts-tasks.ts:21 high-risk-secret-assignment - tokenize assignment
apps/desktop/src/main/database/fts-inbox.ts:22 high-risk-secret-assignment - tokenize assignment
```

(that output is the pre-fix scanner run against the DDL in this PR). The `isSourceCodeReferenceValue` escape hatch never applied either — it bails on quoted values, and the value is `'porter unicode61'`. The `Secret scan` job in `security-ci.yml` runs the same script, so the PR check went red and `--no-verify` was the fastest way past it.

## Pattern before / after

Before — one regex, keyword anywhere inside the key:

```js
/^\s*(?:export\s+)?([A-Z0-9_]*(?:SECRET|TOKEN|...)[A-Z0-9_]*)\s*[:=]\s*["']?([^"'\s#][^#\n]*)/i
```

After — same match, plus a key guard requiring the keyword to stand as its own word. The keyword list is now a single shared source for both the assignment pattern and the guard:

```js
const secretKeywordWordPattern = new RegExp(`(?:^|_)(?:${secretKeywords.join('|')})(?:_|$)`)

function isSecretKeywordKey(key) {
  const words = key
    .replace(/([a-z0-9])([A-Z])/g, '$1_$2')
    .replace(/([A-Z]+)([A-Z][a-z])/g, '$1_$2')
    .toUpperCase()

  return secretKeywordWordPattern.test(words)
}
```

Only the **key** is narrowed. Every value heuristic, the `tokenPatterns` list (GitHub / Slack / AWS / OpenAI / private-key blocks), and the file-scope rules are untouched — no path exclusion, no rule deleted.

The three FTS DDL blocks go back to natural formatting (`tokenize=` on its own line) and the comments explaining the reformat are gone.

## Fixtures — both directions

`scripts/check-staged-secrets.test.mjs`, now wired into the `Secret scan` job (`node --test`, no install needed):

Stops matching: a full `CREATE VIRTUAL TABLE ... tokenize='porter unicode61'` block, `tokenizer: 'porter unicode61'`, `passwordless: 'magic-link-flow'`.

Still matches (`high-risk-secret-assignment`): `API_TOKEN = 'a1b2c3d4-e5f6-7890-abcd-ef1234567890'`, `accessToken: '...'`, `APIToken: '...'`, `API_TOKEN=a1b2c3d4e5f67890abcdef` and `export DB_PASSWORD=hunter2hunter2` in a non-source `.env` path — plus all nine pre-existing "still flags" cases.

Mutation-tested in both directions:

```
=== MUTATION: scanner reverted to origin/main ===
ℹ tests 17
ℹ pass 15
ℹ fail 2
✖ ignores an fts5 tokenize= option in a CREATE VIRTUAL TABLE block
✖ ignores identifiers that only contain a sensitive keyword inside a longer word

=== MUTATION 2: rule narrowed to nothing (isSecretKeywordKey -> false) ===
ℹ tests 17
ℹ pass 8
ℹ fail 9   (every "still flags ..." case)

=== fix in place ===
ℹ tests 17
ℹ pass 17
ℹ fail 0
```

## Local verification

- `node scripts/check-staged-secrets.mjs --changed origin/main` → exit 0
- `node --test scripts/check-staged-secrets.test.mjs` → 17/17
- `eslint` on the changed files → 0 errors; `git diff --check` clean
- `tsc --noEmit -p tsconfig.node.json` (desktop) → exit 0
- `docs:impact --base origin/main --strict` → `covered`; `docs:build` → complete

## Noted, not changed

`privateKey`-style camelCase keys never matched the assignment rule at all (`PRIVATE_KEY` in the keyword list needs the underscore). That is a pre-existing gap in the *widening* direction and out of scope here.

Closes #1143

### DIFF
--- a/.github/workflows/security-ci.yml
+++ b/.github/workflows/security-ci.yml
@@ -45,6 +45,9 @@ jobs:
         with:
           node-version-file: .nvmrc
 
+      - name: Test secret scanner rules
+        run: node --test scripts/check-staged-secrets.test.mjs
+
       - name: Scan changed files for secrets
         env:
           BASE_REF: ${{ github.event_name == 'pull_request' && github.event.pull_request.base.sha || github.event.before }}
--- a/apps/desktop/src/main/database/fts-inbox.ts
+++ b/apps/desktop/src/main/database/fts-inbox.ts
@@ -12,14 +12,14 @@ import type { DataDb } from './client'
  */
 
 export function createFtsInboxTable(db: DataDb): void {
-  // `tokenize=` trails the column above deliberately — see the note in fts.ts.
   db.run(sql`
     CREATE VIRTUAL TABLE IF NOT EXISTS fts_inbox USING fts5(
       id UNINDEXED,
       title,
       content,
       transcription,
-      source_title, tokenize='porter unicode61'
+      source_title,
+      tokenize='porter unicode61'
     )
   `)
 }
--- a/apps/desktop/src/main/database/fts-tasks.ts
+++ b/apps/desktop/src/main/database/fts-tasks.ts
@@ -12,13 +12,13 @@ import type { DataDb } from './client'
  */
 
 export function createFtsTasksTable(db: DataDb): void {
-  // `tokenize=` trails the column above deliberately — see the note in fts.ts.
   db.run(sql`
     CREATE VIRTUAL TABLE IF NOT EXISTS fts_tasks USING fts5(
       id UNINDEXED,
       title,
       description,
-      tags, tokenize='porter unicode61'
+      tags,
+      tokenize='porter unicode61'
     )
   `)
 }
--- a/apps/desktop/src/main/database/fts.ts
+++ b/apps/desktop/src/main/database/fts.ts
@@ -22,17 +22,13 @@ export function createFtsTable(db: IndexDb): void {
   // - content: searchable note body (markdown)
   // - tags: space-separated tags for searching
   // - tokenize: porter stemmer + unicode support for international text
-  //
-  // Keep `tokenize=` trailing the column above rather than starting its own
-  // line: scripts/check-staged-secrets.mjs flags any line-initial assignment
-  // whose key contains "token", so a bare `tokenize=` line fails the Secret
-  // scan CI job on every PR that touches this file.
   db.run(sql`
     CREATE VIRTUAL TABLE IF NOT EXISTS fts_notes USING fts5(
       id UNINDEXED,
       title,
       content,
-      tags, tokenize='porter unicode61'
+      tags,
+      tokenize='porter unicode61'
     )
   `)
 }
--- a/apps/docs/src/contribute/gotchas.md
+++ b/apps/docs/src/contribute/gotchas.md
@@ -202,6 +202,8 @@ GitHub code scanning and the local staged-secret hook are intentionally conserva
 - For generated TypeScript, prefer data tables plus runtime assembly over interpolating dynamic keys into code snippets.
 - In fixtures, avoid object fields named `token`, `secret`, or `apiKey` when the value is runtime data. Use a neutral field name and keep the real header name only at the request boundary.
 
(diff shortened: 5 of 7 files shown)


h4yfans:
Repo-wide before/after on all 5012 tracked files (old scanner from `origin/main` vs the one in this PR):

```
files scanned: 5012
findings old: 10  new: 7
--- suppressed by this PR (old only) ---
  apps/desktop/src/main/database/fts-inbox.ts:22 tokenize assignment
  apps/desktop/src/main/database/fts-tasks.ts:21 tokenize assignment
  apps/desktop/src/main/database/fts.ts:31 tokenize assignment
--- newly flagged (new only) ---
```

The narrowing removes exactly the three fts5 false positives and nothing else — the other seven findings (all in files the changed-file scan never reaches) are untouched, and no new noise appears.

Desktop main-process suite locally: 6033 passed, 1 failed — `telemetry/diagnostics.test.ts` asserts a serialized payload does not contain the substring `secret`, and the worktree path for this branch literally contains `secret-scanner-…`, so the stack trace trips it. Environment-only; CI checks out under `/home/runner/work/memry/memry`.

PR #1360, opened 2026-08-12
fix(tabs): surface quit-time tab-state save failures instead of swallowing them

### LINKED ISSUE #1328: [MEDIUM][tabs] saveSync still swallows quota failures on the beforeunload and Cmd+Q flush paths
Surfaced while fixing epic #987 (memory/CPU audit 2026-08). Not fixed there to keep those PRs surgical.

## What's wrong

PR #1297 (for #1292) made the async save path in `useTabPersistence` handle a rejected write — it catches, keeps `lastSavedRef` on the last value that really reached storage, and raises a toast that distinguishes quota via `isQuotaExceededError` (`apps/desktop/src/renderer/src/contexts/tabs/persistence/hooks.ts:86-107`).

The synchronous last-chance path was not touched. `apps/desktop/src/renderer/src/contexts/tabs/persistence/storage.ts:83`:

```ts
export const saveSync = (state: PersistedTabState): void => {
  try {
    const json = JSON.stringify(state)
    localStorage.setItem(STORAGE_KEY, json)
  } catch (error) {
    log.error('Failed to save tab state synchronously:', error)
  }
}
```

It swallows everything — including `QuotaExceededError`, which `isQuotaExceededError` in the same file (`storage.ts:25`) already knows how to identify. Both callers are the last-write paths:

- `hooks.ts:59` — the Cmd+Q flush registry callback, which then logs `flushed tab state via registry` whether or not anything was written
- `hooks.ts:124` — the `beforeunload` handler

## Why it matters

On the quit path, a full-quota origin means the final write silently fails while the log claims a flush happened, and whatever stale tab state is already in `localStorage` is what the user gets restored on next launch. A toast is useless mid-quit — nobody will see it — so the honest consequence is narrower than #1292's: the failure is unobservable rather than user-facing, and there is no signal beyond one renderer log line for anyone diagnosing "my tabs came back wrong after quitting".

## Suggested direction

Give `saveSync` a return value (or let it throw) so callers can tell success from failure, and have the flush-registry callback stop logging a successful flush when nothing was written; classifying via the existing `isQuotaExceededError` would let the log line say *why*. Whether it is worth surfacing to the user at all on the quit path is the open question — a caller that still has an async window (`beforeunload` sometimes does) might do something more useful than logging.

## Confirmed on

`origin/main` @ c0754c9b7 on 2026-08-12.

h4yfans:
Closes #1328. Part of epic #987 (memory/CPU audit 2026-08).

## What was wrong

`saveSync` — the last-chance write on both quit paths — returned `void` and swallowed every failure, quota included (`apps/desktop/src/renderer/src/contexts/tabs/persistence/storage.ts:83` on `origin/main`):

```ts
export const saveSync = (state: PersistedTabState): void => {
  try {
    localStorage.setItem(STORAGE_KEY, JSON.stringify(state))
  } catch (error) {
    log.error('Failed to save tab state synchronously:', error)
  }
}
```

Its callers therefore could not tell a write from a no-op. The Cmd+Q flush-registry callback (`hooks.ts:59`) logged `flushed tab state via registry` unconditionally, and the `beforeunload` handler (`hooks.ts:124`) logged nothing at all. On a full origin the final write is refused, the stale payload already in `localStorage` is what gets restored next launch, and the only trace is a log line contradicted by the "flushed" line right after it.

PR #1297 fixed the async sibling (catch + toast + `lastSavedRef` kept on the last value that really landed). This is the synchronous half.

## What changed

- `saveSync` now returns `{ ok: true } | { ok: false; reason: 'quota' | 'error'; at: number }`, classifying via the `isQuotaExceededError` helper that already lived in the same file. It still never throws — both callers run while the window is closing, where an exception would be a far worse regression than the lost write.
- On failure it writes a tiny marker key (`memry_tab_state_save_failed`) and clears that marker on any successful sync save. The marker write is itself best-effort and wrapped, so a truly full origin degrades to a log line instead of an exception.
- Both quit callers now log the truth: `failed to flush tab state via registry` / `failed to save tab state on unload` with the reason, and the success line only when the write actually landed.
- `useSessionRestore` consumes the marker on the next launch and raises one toast — *"Your last session was not saved when Memry closed"* — with a quota-specific or generic description. Consuming is one-shot, so a single failed quit cannot nag on every later launch.
- Three English strings under `tabs.persistence.*`; docs section added to `apps/docs/src/user-guide/tabs-split-view.md` next to the one #1297 added.

Deliberately **not** done:

- **No toast at quit time.** The window is already going away; nobody would read it. The next launch is the first moment the message is worth anything, which is exactly when the user is looking at a layout that is older than the one they left.
- **No retry with a trimmed payload.** Dropping tabs to make the write fit is the same data loss with less evidence, and it would silently change what "restore session" means. And no `removeItem`-then-retry either: Chromium accounts for a replaced value when it checks quota, so freeing our own key first buys nothing.
- **No IPC to the main process.** A fire-and-forget `invoke` during `beforeunload` is not reliably delivered, and it would drag a contract change into a renderer-local problem.

## How it was proven

`persistence.test.tsx`, renderer project:

- **Before** (source reverted to `origin/main`, tests kept): **7 failed | 9 passed (16)** — `expected undefined to deeply equal { ok: false, reason: 'quota', … }`, `consumeSyncSaveFailure is not a function`, `expected "vi.fn()" to not be called with arguments: [ 'flushed tab state via registry' ]`, `expected "vi.fn()" to be called 1 times, but got 0 times`.
- **Before, caller-side only** (only `hooks.ts` reverted, so the new storage API is present): **2 failed | 14 passed (16)** — the pre-fix flush callback still claims `flushed tab state via registry` on a refused write, and the next launch never warns.
- **After**: **16 passed (16)**.

New coverage: refused sync save reports `{ ok: false, reason: 'quota' }` and leaves a marker; the marker is consumed once; a later successful save clears it; a corrupt/foreign marker is ignored; a **normal quit still writes the state and logs the flush** (`expect(mocks.logger.error).not.toHaveBeenCalled()`); neither quit path throws when the write is refused and the stale payload is left untouched; the next launch warns once, with the quota wording for a quota failure and the generic wording otherwise.

## Verification

```
./node_modules/.bin/vitest run --config apps/desktop/config/vitest.config.ts --project renderer \
  apps/desktop/src/renderer/src/contexts/tabs/persistence/persistence.test.tsx
  → Test Files 1 passed (1) | Tests 16 passed (16)

pnpm --filter @memry/desktop typecheck:web   → clean
pnpm --filter @memry/desktop i18n:check      → ok: i18n check passed
./node_modules/.bin/eslint apps/desktop/src/renderer/src/contexts/tabs/persistence/  → 0 problems
packages/i18n vitest run                     → 11 files, 56 tests passed
pnpm docs:impact --base origin/main --strict → docs changed on this branch
pnpm docs:build                              → build complete
git diff --check                             → clean
```

## Backward compatibility

Additive only: the tab-state payload, `STORAGE_KEY`, and the persisted schema are untouched, the new marker lives under its own key that older builds simply ignore, and an install upgrading with no marker present behaves exactly as before.

### DIFF
--- a/apps/desktop/src/renderer/src/contexts/tabs/persistence/hooks.ts
+++ b/apps/desktop/src/renderer/src/contexts/tabs/persistence/hooks.ts
@@ -10,7 +10,12 @@ import { toast } from 'sonner'
 import { useTabs } from '@/contexts/tabs'
 import type { TabSystemState } from '@/contexts/tabs/types'
 import { extractErrorMessage } from '@/lib/ipc-error'
-import { getDefaultStorage, isQuotaExceededError, saveSync } from './storage'
+import {
+  consumeSyncSaveFailure,
+  getDefaultStorage,
+  isQuotaExceededError,
+  saveSync
+} from './storage'
 import { serializeTabState, deserializeTabState, extractPinnedTabs } from './serialization'
 import type { TabStorage } from './types'
 import { registerPendingSave, unregisterPendingSave } from '@/lib/save-registry'
@@ -20,6 +25,7 @@ import { useFeatureFlags } from '@/hooks/use-feature-flags'
 const log = createLogger('TabPersistence:Hooks')
 const FLUSH_REGISTRY_KEY = 'tab-state'
 const SAVE_FAILURE_TOAST_ID = 'tab-state-save-failed'
+const LAST_SESSION_TOAST_ID = 'tab-state-last-session-not-saved'
 
 // =============================================================================
 // AUTO-SAVE HOOK
@@ -56,7 +62,14 @@ export const useTabPersistence = (options: UseTabPersistenceOptions = {}): void 
 
     registerPendingSave(FLUSH_REGISTRY_KEY, () => {
       const serialized = serializeTabState(stateRef.current)
-      saveSync(serialized)
+      const result = saveSync(serialized)
+      if (!result.ok) {
+        // Claiming a flush that never happened is what made this invisible:
+        // whoever diagnoses "my tabs came back wrong after quitting" needs the
+        // log to say the write was refused, and why.
+        log.error('failed to flush tab state via registry', { reason: result.reason })
+        return
+      }
       log.info('flushed tab state via registry')
     })
 
@@ -136,7 +149,10 @@ export const useTabPersistence = (options: UseTabPersistenceOptions = {}): void 
 
     const handleBeforeUnload = (): void => {
       const serialized = serializeTabState(stateRef.current)
-      saveSync(serialized)
+      const result = saveSync(serialized)
+      if (!result.ok) {
+        log.error('failed to save tab state on unload', { reason: result.reason })
+      }
     }
 
     window.addEventListener('beforeunload', handleBeforeUnload)
@@ -219,6 +235,23 @@ export const useSessionRestore = (
         }
       } else {
         log.info('no persisted tab state found')
+      }
+
+      // The quit-time write is the one save the user can never be told about
+      // while it fails — the window is already closing. If it left a marker,
+      // this launch is the first moment the explanation is worth anything:
+      // whatever just got restored is older than what they closed the app with.
+      const saveFailure = consumeSyncSaveFailure()
+      if (saveFailure) {
+        log.warn('previous session was not saved on exit', { reason: saveFailure.reason })
+        const t = getI18n().getFixedT(null, 'common')
+        toast.warning(t('tabs.persistence.lastSessionNotSaved'), {
+          id: LAST_SESSION_TOAST_ID,
+          description:
+            saveFailure.reason === 'quota'
+              ? t('tabs.persistence.lastSessionNotSavedQuota')
+              : t('tabs.persistence.lastSessionNotSavedGeneric')
+        })
       }
 
       hasRestoredRef.current = true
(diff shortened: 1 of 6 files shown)


h4yfans:
Rebased onto `origin/main` after #1367 landed.

Conflicts resolved: `index.ts` (kept #1367's `indexedDBAdapter` removal, kept my `SyncSaveFailure`/`SyncSaveResult` type export) and `persistence.test.tsx` (kept #1367's deletion of the IndexedDB test + `installIndexedDbMock` harness and its new dedupe test, kept my six). `hooks.ts` and `storage.ts` auto-merged; nothing here used `toError` or anything else #1367 deleted. Both docs additions to `tabs-split-view.md` coexist.

**Does a skipped write clear the failure marker? No — and it can't.** #1367's short-circuit is the `return` inside the *debounced async* callback (`hooks.ts:109`), which gates `storage.save` only. The marker is written and cleared exclusively by `saveSync`, and both quit callers invoke it unconditionally — #1367's own comment says as much ('both quit paths write unconditionally regardless of this guard'). So a skipped write cannot touch the marker, which is the correct semantics: the marker records that the last quit-time write never landed, and skipping a write does not land one. If anything the dedupe strengthens the case, since the quit-time write is now more often the only write in a quiet session.

Fresh numbers on the rebased tree:
- before, full source reverted to `origin/main`: **7 failed | 9 passed (16)**
- before, only `hooks.ts` reverted: **2 failed | 14 passed (16)**
- after: **16 passed (16)**; whole tabs directory **11 files, 115 passed**
- `typecheck:web` clean, eslint on the persistence dir clean, `docs:impact --base origin/main --strict` → docs changed on this branch, `git diff --check` clean

PR #1370, opened 2026-08-12
fix(tooling): serialise the Electron binary install and swap it in atomically

### LINKED ISSUE #1343: [LOW][tooling] Fresh-worktree first vitest run can fail suites with "Electron failed to install correctly" — the Electron install is an unlocked destructive check-then-act
Surfaced while fixing epic #987 (memory/CPU audit 2026-08). Not fixed there to keep those PRs surgical.

## What's wrong

In a fresh worktree, the **first** `vitest` run can fail several main-project suites at import time with `Error: Electron failed to install correctly` and, alongside it, `unzip` errors of the form `failed to create directory .../dist/Electron.app: File exists`. Re-running the suite is green with no intervention. Observed today across several agents working in parallel worktrees; it cost real time because the failures read as a logic regression, not as an environment problem.

The mechanism that makes this possible is in the tooling, and is static:

**The Electron install is a destructive, unlocked, check-then-act.** `apps/desktop/scripts/ensure-native.sh:38-57` tests for the binary and then installs:

```sh
has_electron_binary() {
  if [ ! -f "$ELECTRON_DIR/path.txt" ]; then
    return 1
  fi
  ...
  [ -f "$ELECTRON_DIR/dist/$electron_path" ]
}

install_electron_binary() {
  node "$SCRIPT_DIR/install-electron-binary.cjs" "$ELECTRON_DIR"
}
```

There is no lock between the test and the install. `apps/desktop/scripts/install-electron-binary.cjs` `main()` then wipes the shared target before extracting into it:

```js
  fs.rmSync(distDir, { recursive: true, force: true })
  fs.rmSync(pathFile, { force: true })
  ...
    extractArtifact(zipPath, distDir)   // childProcess.execFileSync('unzip', ['-q', zipPath, '-d', distDir])
    ...
    fs.writeFileSync(pathFile, platformPath)
```

Two invocations racing on the same `dist/` are exactly what produces `unzip`'s "File exists" on a half-removed tree; and because `path.txt` is written **last**, any interleaving leaves a window where `dist/` is present but the marker is not.

**That window is what surfaces as a test failure.** `node_modules/electron/index.js` throws `Electron failed to install correctly, please delete node_modules/electron and try installing again` at `require()` time when `path.txt`/`dist` is missing. Any vitest file that loads real `electron` — directly or transitively via `electron-log`, which is an external runtime dependency (`apps/desktop/src/main/runtime-dependencies.test.ts:21`) — then fails during module setup, so the whole file is reported as failed rather than a single test. Files that stub electron pass, which is why the failure set looks arbitrary.

This is the same failure family already known internally as the "fresh worktree electron marker" problem (`dist/` extracted but `path.txt` missing or wrong), reached by a different route.

## Why it matters

Purely a DX cost, nothing user-visible. But the symptom is maximally misleading: ~dozens of main-project files fail fast at import with errors like `db.select is not a function` next to the Electron message, so it reads as a broken branch. Every agent or contributor who hits it burns time bisecting a regression that does not exist, and the "just run it again" fix is only obvious once you already know.

## Suggested direction

Serialise the install: take an exclusive lock (an `O_EXCL` marker file, or `flock` where available) around `install_electron_binary` in `ensure-native.sh`, and make `install-electron-binary.cjs` extract into a sibling temp directory and `rename()` it into place so `dist/` is never observed half-written. Writing `path.txt` before the rename would also close the "dist present, marker missing" window. Worth confirming first *what* is invoking the installer concurrently in the reported scenario — `pretest` runs `ensure-native.sh node`, which does not install the Electron binary, so the second racer may be a parallel `pnpm dev` / `pretest:e2e` / postinstall rather than a second test file; the reproduction should be nailed down before the fix is sized.

## Confirmed on

`origin/main` @ c0754c9b7 on 2026-08-12. The concurrency hole in `ensure-native.sh` / `install-electron-binary.cjs` was read on that sha; the intermittent test failure itself was reported from a live session and was not re-reproduced for this issue.

h4yfans:
Closes #1343. Part of epic #987 (memory/CPU audit 2026-08).

## What was wrong

`apps/desktop/scripts/install-electron-binary.cjs` wiped the shared install before it had a replacement, and nothing serialised concurrent runs:

```js
fs.rmSync(distDir, { recursive: true, force: true })
fs.rmSync(pathFile, { force: true })
// ...seconds of download + unzip...
extractArtifact(zipPath, distDir)
fs.writeFileSync(pathFile, platformPath)   // written last
```

`node_modules/electron/index.js` throws `Electron failed to install correctly` at `require()` time whenever `path.txt` or `dist/` is missing. So for the whole download+extract window, any Vitest file that loads real `electron` — directly or transitively via `electron-log` — dies during module setup and the whole file is reported failed. Files that stub `electron` pass, which is why the failure set looks arbitrary and reads as a logic regression.

**The racer, since the issue asked for it to be nailed down first.** The filer's guess that worktrees share one Electron cache is wrong: every worktree has its own `node_modules/.pnpm/electron@43.1.1/node_modules/electron/dist` (verified — distinct inodes, link count 1). The real second entrant is in-process and in-repo:

- `apps/desktop/tests/e2e/utils/electron-lifecycle.ts:49-89` — `getElectronExecutablePath()` is itself an unlocked check-then-act that shells out to this same installer when `path.txt` is missing, and it runs **per Playwright worker** (`apps/desktop/config/playwright.config.ts:24`, `workers: 2` locally). One `pnpm test:e2e` is therefore enough to put two installers on the same `dist/`.
- `ensure-native.sh` reaches the same script from `predev` / `prebuild` / `pretest:e2e` / `rebuild:electron`, so a parallel `pnpm dev` is a second route.

Confirmed by reproduction rather than by reading. A harness drives the real script offline (fixture zip served by a `curl` shim earlier on `PATH`, real `unzip`), against `origin/main`:

- one reinstall, polled by a concurrent reader every 5 ms → **108 of 119 samples saw a broken install**
- two concurrent installers → **both exit 1**, and `dist/` is left permanently unusable (`unzip: ... File exists`, `write error`)

## What changed

One file does the work — `install-electron-binary.cjs`:

- **Serialised.** An exclusive `O_EXCL` lock file next to the `electron` package. The loser waits, then re-checks under the lock and no-ops instead of re-downloading. Stale locks are reclaimed: the holder's pid is probed with `process.kill(pid, 0)`, with a 10-minute age backstop for a live-but-wedged holder, so it cannot deadlock.
- **Atomic.** The archive extracts into a `.dist-staging-<pid>` sibling (same filesystem — `os.tmpdir()` can be a different device, EXDEV), is validated there before the live `dist/` is touched at all, then swapped in with two renames. `path.txt` is the commit marker: written last, and never deleted. A reader now sees either the previous install or the new one.
- Scratch dirs left by a killed run are swept under the lock; the pre-lock temp dir is cleaned on the early-return path too.

**Deliberately not done:**

- **No `flock` in `ensure-native.sh`.** macOS ships no `flock` binary, and once the installer is itself locked and idempotent the shell-level check-then-act is harmless — the second entrant simply waits and no-ops. Same reasoning for `electron-lifecycle.ts`: both workers may spawn the installer, one installs, the other returns the finished path. Putting the lock at the single choke point both callers already go through keeps this to one file and works on every platform.
- **`path.txt` is not written before the swap**, contrary to the issue's suggestion. Readers check `path.txt` *and* `dist/<path>`, so publishing the marker early would just move the broken window rather than close it.
- The pre-existing `dist/` absent after a fresh `pnpm install` (see below) is out of scope.

## How it was proven

New `apps/desktop/scripts/install-electron-binary.test.cjs` — 4 tests, fully offline, driving the real script through a `curl` shim and real `unzip`/renames.

Per the brief, source-only revert (`git checkout origin/main -- apps/desktop/scripts/install-electron-binary.cjs`, test file untouched):

| | tests | pass | fail |
|---|---|---|---|
| before (`origin/main` source) | 4 | **0** | **4** |
| after (fix restored) | 4 | **4** | **0** |

The headline before-failure: `reader saw a broken install in 121/168 samples`.

Also mutation-tested rather than assumed: removing just the early-return scratch cleanup turns the concurrency test red with `an install leaked a scratch directory` (3 pass / 1 fail), then green again on restore.

## Verification

```
node --test apps/desktop/scripts/install-electron-binary.test.cjs
  → tests 4, pass 4, fail 0

pnpm exec eslint apps/desktop/scripts/install-electron-binary{,.test}.cjs
  → 0 errors (both files match an ignore pattern)

pnpm exec prettier --check <changed files>   → All matched files use Prettier code style
pnpm docs:impact --base origin/main --strict → docs impact: covered
pnpm docs:build                              → build complete in 3.54s
git diff --check                             → clean
```

No TypeScript changed, so `typecheck:web` / `typecheck:node` are not applicable. Wired into CI next to the existing tooling-script tests (`.github/workflows/desktop-ci.yml`), and it skips itself where `zip`/`unzip`/POSIX `sh` are unavailable.

**Not verified here:** a real Electron download. This sandbox cannot reach the GitHub release mirror (`curl: (56) Connection died`) — `origin/main`'s script fails identically on the same box, so that is environmental and not a regression. The swap is a directory `rename`, which does not touch contents, permissions, or the symlinks inside `Electron.app`, but a real `pnpm dev` on an unrestricted network is worth one manual pass before this leaves draft.

## Backward compatibility

Tooling only — no runtime, IPC, schema, sync, or settings surface is touched, and the installed layout (`dist/` + `path.txt`) is byte-for-byte what it was, so existing worktrees and CI caches keep working unchanged.

### DIFF
--- a/.github/workflows/desktop-ci.yml
+++ b/.github/workflows/desktop-ci.yml
@@ -94,6 +94,12 @@ jobs:
       - name: Test e2e keychain cleanup tooling
         run: node --test apps/desktop/scripts/purge-e2e-keychain.test.mjs
 
+      # Guards the install lock + atomic swap that keep a concurrent Electron
+      # install from breaking unrelated suites at `require('electron')` time.
+      # Runs fully offline against a fixture zip.
+      - name: Test Electron binary installer concurrency
+        run: node --test apps/desktop/scripts/install-electron-binary.test.cjs
+
   unit:
     name: Unit & integration tests
     runs-on: ubuntu-latest
--- a/apps/desktop/scripts/install-electron-binary.cjs
+++ b/apps/desktop/scripts/install-electron-binary.cjs
@@ -9,6 +9,17 @@ const path = require('path')
 const electronDir = process.argv[2]
 const officialMirror = 'https://github.com/electron/electron/releases/download/'
 
+// Several processes can reach this script at once on one machine: `ensure-native.sh`
+// (predev / prebuild / pretest:e2e / rebuild:electron) and every Playwright worker,
+// which falls back to installing when `path.txt` is missing (see
+// tests/e2e/utils/electron-lifecycle.ts, `workers: 2` locally). They all target the
+// same `dist/`, so the install is serialised with a lock and swapped into place
+// atomically — otherwise concurrent `unzip`s collide and readers of the package
+// (`require('electron')` throws without `path.txt` + `dist/`) observe a half-written
+// tree and fail with "Electron failed to install correctly".
+const LOCK_POLL_MS = 250
+const LOCK_STALE_MS = 10 * 60 * 1000
+
 if (!electronDir) {
   console.error('Usage: install-electron-binary.cjs <electron-package-dir>')
   process.exit(2)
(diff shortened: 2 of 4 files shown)


h4yfans:
### Real-download verification (the caveat in the PR description is now closed)

Ran on macOS arm64 with unrestricted network access to the GitHub release mirror, against this branch's installer, in the `electron-install-lock` worktree — whose `electron` package started with **no `dist/` and no `path.txt`** (the second, race-free cause noted at the end of the PR body).

```
$ curl -sS -o /dev/null -w "%{http_code}" -L \
    https://github.com/electron/electron/releases/download/v43.1.1/SHASUMS256.txt
200

$ node apps/desktop/scripts/install-electron-binary.cjs "$PWD/node_modules/.pnpm/electron@43.1.1/node_modules/electron"
[electron] installing 43.1.1 for darwin-arm64 from https://github.com/electron/electron/releases/download/
[electron] installed  43.1.1 for darwin-arm64: .../dist/Electron.app/Contents/MacOS/Electron
        6.35s total
```

Result: `dist/` present (305 MB), `path.txt` = `Electron.app/Contents/MacOS/Electron`, written last as designed.

**The staging → rename swap does not break the app bundle** — the extracted binary runs:

```
$ .../dist/Electron.app/Contents/MacOS/Electron --version
v43.1.1
```

**Sequential idempotency** — 3 further runs: `exit=0`, `exit=0`, `exit=0`, binary still `v43.1.1`.

**Concurrent installers — the case this PR exists for** — 3 processes launched simultaneously against the same `dist/`:

```
$ for i in 1 2 3; do node .../install-electron-binary.cjs "$d" & done; wait
all exit 0
$ .../Electron --version
v43.1.1
```

On `origin/main` the same shape is what the PR reproduces as both installers exiting 1 and leaving `dist/` permanently unusable. Here all three succeed and the install is intact afterwards.

Verified by the coordinating session, not by the authoring agent (whose sandbox could not reach the mirror).

Issue #808, opened 2026-07-20
[Feature] Rename an attachment's title from the inbox (image/PDF)


h4yfans:
**Requested by:** Aurelie Kabore — v2026.718.3
**Area:** Inbox
**Verdict:** Valid feature request (backend already supports it)

## Ask
"Would you might making editing title of attachment in inbox work? Either right clicking to rename or when that right panel is opened?"

## Current state
The detail panel shows an editable title input **only** for `type:'voice'` (`inbox-detail-panel.tsx:447-461`); for `image`/`pdf` the title isn't shown (`:463-466`). No rename entry in the inbox row context menu (`inbox/inbox-list.tsx`). The mutation already exists and handles title: `useUpdateInboxItem({id, title})` (`inbox-detail-panel.tsx:452`).

## Direction
Reuse `useUpdateInboxItem({id,title})` — render the editable title input for image/PDF (not just voice), and/or add a right-click "Rename" in `inbox-list.tsx`.

---
Part of #795

Issue #122, opened 2026-04-02
feat: activity log / history per task


h4yfans:
> **Rewritten 2026-08-13.** The original body (2026-04-02) prescribed IPC-handler middleware as the logging hook — that is wrong for today's architecture and would have missed most task writers. See the comment thread for what changed.

## Summary

Per-task audit trail — who changed what, when, and from which device — synced across devices and surfaced in the task detail drawer.

Two jobs, and the second one is the reason this is worth building:

1. **User trust.** "Why is this task's due date different from what I set?"
2. **Multi-device sync debugging.** Today, when two devices edit the same field concurrently, `field-merge.ts` picks a winner and the loser's edit vanishes with no user-visible trace. `conflictedFields` is computed and then only written to a log line (`item-handlers/task-handler.ts:141`); the renderer receives `changes: {}` (`:187`). Surfacing that is in scope here.

## Decisions (locked)

| Decision | Choice |
|---|---|
| Sync | **Full synced item type** — new `task_activity` `SyncItemType`, not local-only |
| Scope | **Tasks only** — no projects, no generic `item_activity` table |
| Conflicts | **In scope** — emit `superseded` rows when a local edit loses a merge |
| Storage | data DB (`packages/db-schema/src/data-schema.ts`), never index DB |

---

## Where to hook — the choke point

**This replaces the original "wrap IPC handlers" plan.** IPC handlers are not the choke point; the domain publisher is. IPC middleware would miss MCP/agent writes, the Todoist/TickTick importers, markdown-checkbox reconcile, and note runtime effects — all of which mutate tasks through the domain but never through `tasks-handlers.ts`.

Hook **`apps/desktop/src/main/tasks/publisher.ts`** → `createTasksPublisher()` (line 32). Six callbacks cover every task mutation, because the domain already fans bulk operations out into per-id events:

| callback | line | note |
|---|---|---|
| `taskCreated` | 34 | |
| `taskUpdated` | 45 | already receives `changedFields` |
| `taskDeleted` | 57 | already receives `snapshot` |
| `taskCompleted` | 68 | |
| `taskMoved` | 78 | |
| `taskReordered` | 82 | see noise rules below |

Publisher is constructed in `apps/desktop/src/main/tasks/domain.ts`. Everything below flows through it for free:

- `src/main/ipc/tasks-handlers.ts` (40+ channels)
- `src/main/agent/mcp/tools/handles-adapter.ts` (Agent Chat / MCP)
[...]

PR #1385, opened 2026-08-12
feat(tasks): per-task activity log (#122)

h4yfans:
## Summary
Per-task audit trail, closing #122: a `task_activity` synced item type (hand-written `0049` migration), a write path hooked to `createTasksPublisher()` — the one choke point every task mutation reaches, including Agent Chat/MCP, the importers, markdown checkbox reconcile and the two raw-write bypasses (inbox filing, Google Calendar writeback) — and a three-row preview in the task detail drawer with a filterable Sheet behind "Show all". Rows are append-only, so apply is insert-if-absent and retention is a shared 90-day age rule enforced on apply as well as on read; `description` is stored as a length delta and never as the body. Concurrent edits that lost a merge become `superseded` rows whose id derives from the *merged* field clock rather than the winning device — clock merge is commutative, so both sides of a mirrored conflict mint the same id and collapse, where keying off the winner would never converge. The UI never names a device: "You" or "Another device".

Three independent reviews (sync, write path, UI) ran against the first two commits; `d5156a777` fixes what they found — see that commit message for the list.

## Release note
Tasks now keep an activity log. Open a task and scroll to Activity to see what changed, when, and whether an edit was replaced by another device.

## Test plan
- `pnpm typecheck` (runs `check:contracts`, `check:architecture`, `ipc:check`) — green
- `pnpm lint` — 0 errors
- `pnpm --filter @memry/desktop i18n:check` — passed
- `pnpm test:sync-server` — 936 passing
- Full desktop suite — 15702 passing, incl. new coverage for the sync handler, the write path (real publisher against a real SQLite DB, not a mocked writer), the read query, and the drawer section
- `pnpm docs:impact --base origin/main --strict` and `pnpm docs:build` — green

**Not yet run:** the two-profile manual sync check (`dev:a` / `dev:b`) — concurrent edit of one field should produce exactly one `superseded` row across both devices with no re-sync loop.

## Both red checks are inherited from `main`, not from this diff
- **Unit & integration tests** — the same 3 tests fail locally on this branch's base, all from `24e6b6fce` (#808 inbox rename): `component-major-surfaces.test.tsx > ListView` (`useUpdateInboxItem` missing from the `@/hooks/use-inbox` mock) and two `filing.test.ts` cases asserting the old filed-attachment path. This diff touches no renderer inbox or filing code.
- **Secret scan** — flags `apps/desktop/src/main/sync/setup-token.ts:51,65`, a file that does not exist on this branch. It arrived on `main` in `0bb966246` (#1382), and PR runs scan the merge commit, so every PR based before that commit inherits it. Both hits are the bare declaration `let secretKey: Uint8Array | undefined` — a type annotation with no value, matched by the same `secretAssignmentPattern` wart that already false-positives on SQL `tokenize=`.


h4yfans:
Follow-up commit `9e223cd2d` — the feed was reporting fields nobody edited.

The tasks page sends the whole updatable key set on every edit and leaves the untouched fields `undefined`, so the domain differ read each of them as "cleared": a status-only change also logged title, description, due date, due time, project and repeat. Nothing had actually changed (drizzle omits `undefined` from `.set()`), but `changedFields` also feeds `syncTaskUpdate`, so the edit bumped the field clocks of fields it never touched and could win a field-level merge against a peer that really did edit them.

`undefined` now means "not supplied" and is dropped before the diff; `null` still means "clear this field". The Google Calendar writeback had the same defect in its own differ, and a device with no registered id now recognizes its own offline-stamped rows instead of labelling them as another device's.

Four tests, each verified to fail without the fix.

Docs gate waived (`MEMRY_DOCS_IMPACT_SKIP=1`): this restores the behaviour `apps/docs/src/user-guide/tasks/activity-log.md` already documents — "Each entry names the field that changed", and "**You** for this device". No user-facing statement changes.

Issue #1468, opened 2026-08-15
EPIC: large files and large notes — stop the freeze, stop the silent sync failure


h4yfans:
Umbrella for two related failures that share one root gap: **nothing in the note pipeline has a size ceiling, and nothing tells the user when a size limit is hit.**

Close this epic when every child below is closed.

## How this started

A 17.8 MB log dump pasted into the vault from Finder froze the whole app. Reproduced on three installs — two signed-in devices and one signed-out. A 1.9 MB dump caused a shorter freeze that recovered. Investigating it surfaced a second, unrelated-looking failure with the same shape: a note that grows past the sync ceiling stops syncing and nobody is told.

## Root cause of the freeze

The vault watcher's markdown-add handler calls the note-create runtime effect, which initialises the note's CRDT doc, which seeds it by running the BlockNote markdown parser over the entire file on the main process with no yield point. That call is made just before the renderer event that paints the sidebar row and is not awaited — so the row appears, then the parse locks the main process. The report was "it freezes the moment I see it in the sidebar", which is exactly what the call ordering predicts.

Measured against the real files. The metadata regexes and content hash cost ~20 ms for 1.9 MB, and the FTS5 insert costs 93 ms for 17 MB — **neither is the bottleneck**. The parse is, and its cost tracks **single-block size**, not file size:

```
1.81 MB, blank-line separated (124 blocks)  ->    477 ms
1.81 MB, as-is                (1 block)     ->  6 594 ms      14x
2.73 MB, as-is                (2 blocks)    ->  8 936 ms
```

Log dumps have no blank lines, so the whole file parses as one paragraph holding millions of inline nodes. 17.8 MB extrapolates to roughly ten minutes.

## Ceilings that bound the design

```
V8 max string length     536,870,888 chars (~512 MB)  -> reading a bigger file as one string throws
SQLite MAX_LENGTH        2,147,483,645 bytes (~2 GB)
client encrypt cap       5 MiB on snapshot x 1.37     -> ~3.7 MB of markdown
server body cap          8 MiB on /sync/*
plan maxFileSize         plus 5 MiB, pro/believer 200 MiB — blob routes only, never note bodies
```

## Decisions

- Two classes of vault file. **Note class** — editable, BlockNote, CRDT-seeded, synced. **Large-file class** — read-only streaming viewer, no Y.Doc, no sync, visible badge.
[...]

Issue #1459, opened 2026-08-15
Oversized markdown must never become a CRDT note


h4yfans:
## Parent

- #1468

## What to build

Dropping a large markdown file into the vault from Finder freezes the whole app — at paste time, before the user has opened anything. Reported across three installs: two signed-in devices and a signed-out one.

Root cause: the vault watcher's markdown-add handler calls the note-create runtime effect, which initialises the note's CRDT doc, which seeds it by running the BlockNote markdown parser over the entire file on the main process with no yield point. That call is made just before the renderer event that paints the sidebar row and is not awaited, so the row appears and the parse then locks the main process. The report was "it freezes the moment I see it in the sidebar" — exactly what the call ordering predicts.

Measured against the real files. The metadata regexes plus content hash cost ~20 ms for 1.9 MB, and the FTS5 insert costs 93 ms for 17 MB — neither is the problem. The parse is, and its cost tracks **single-block size**, not file size:

```
1.81 MB, blank-line separated (124 blocks)  ->    477 ms
1.81 MB, as-is                (1 block)     ->  6 594 ms      14x
2.73 MB, as-is                (2 blocks)    ->  8 936 ms
```

Log dumps contain no blank lines, so the whole file parses as one paragraph holding millions of inline nodes. A 17.8 MB file extrapolates to roughly ten minutes.

Build a pure classifier in `packages/shared` that decides, from file size and largest blank-line-separated block, whether a markdown file is **note class** (editable, BlockNote, CRDT-seeded, synced — today's behaviour) or **large-file class**. Have both the watcher ingest path and the CRDT seed path consult it. Large-file class never gets a Y.Doc. Opening one shows a placeholder explaining it cannot be opened as a note — a real message, not a hang and not an empty editor.

Starting thresholds, both must hold for note class:

- file size <= 2 MB
- largest blank-line-separated block <= 128 KB

A fixed byte ceiling alone misclassifies in both directions: it rejects a well-formed 3 MB note that parses fine, and accepts a 900 KB log dump that does not. The pair is the point. The numbers get calibrated in a later ticket.

Do not widen the `FileType` union. It is persisted in the note cache's file-type column and appears in projection types and IPC contracts; existing rows say `markdown` and must keep working. Note class vs large-file class is a separate axis.

The seed guard must apply regardless of when the note was created, so oversized notes already sitting in existing vaults stop freezing on open too.

## Acceptance criteria

[...]

Issue #1463, opened 2026-08-15
Calibrate the note-class thresholds against a real note corpus


h4yfans:
## Parent

- #1468

## What to build

The note-class thresholds shipped in #1459 — 2 MB file size and 128 KB largest blank-line-separated block — rest on a single measurement: 1.81 MB across 124 blocks parsing in 477 ms, against 6 594 ms for the same bytes as one block. That is enough to prove the shape of the curve but not enough to pick the numbers.

Measure the BlockNote markdown parse across a corpus of real notes, spanning small prose notes, long structured documents, imported vaults and pathological machine-generated files. Derive the two thresholds from a stated parse budget rather than from one data point, and document the budget, the corpus and the resulting numbers so the next person does not have to re-derive them.

The risk of the current values is low — 2 MB is roughly 300k words of prose, well beyond any real note — so this is calibration, not a correctness fix.

## Acceptance criteria

- [ ] Parse timings captured across a corpus covering prose, structured, imported and machine-generated markdown
- [ ] A stated parse budget, with the reasoning for that budget
- [ ] Both thresholds derived from the budget and updated if the measurements call for it
- [ ] Corpus, method and resulting numbers documented alongside the classifier
- [ ] Classifier table tests extended with any newly discovered edge shapes

## Blocked by

- #1459

h4yfans:
Done in #1478 (merged as `54f5b3bcc`).

**Budget:** a note's first open may block the main process for at most **1 s** — it is a full main-process stall with no yield point, macOS beachballs at ~2 s, the measuring machine is an Apple Silicon laptop while users run hardware 2–4x slower, and the cost is paid once per note so a tight budget is cheap to hold.

**Result:** `NOTE_MAX_BYTES` 2 MB → **1 MB**; `NOTE_MAX_BLOCK_BYTES` **confirmed at 128 KB**.

Measured through `markdownToBlocks` across a generated corpus (prose, structured docs, imported Obsidian vault notes, Roam/Bear tight outlines, log dumps, wide tables, minified JSON):

- Well-formed prose is **450 ms/MB and linear**; an imported vault note is 830 ms/MB; a table-dense document ~1 400 ms/MB and superlinear. At 2 MB that is 891 / 1 654 / 2 990 ms — two of three realistic shapes over budget. At 1 MB: 429 / 830 / ~1 400 ms.
[...]

Issue #1461, opened 2026-08-15
Large-file class must not sync, on either side


h4yfans:
## Parent

- #1468

## What to build

One of the two devices that froze never touched Finder. It received the note over sync — its log shows the CRDT writeback creating the note from sync — and then paid the same BlockNote parse locally. The snapshot push from the sending device had been rejected at the 5 MB encrypt cap (see #1445), so the body arrived as incremental CRDT updates instead, which are capped at 256 KB merged and 512 KB per flush and therefore pass freely.

An ingest-side guard alone would not have saved that device. The classification from #1459 has to apply on both sides of sync.

Sending side: a large-file-class file produces no sync item at all, so it does not appear in other devices' sidebars. Showing a row that cannot be opened is worse than not showing it. Syncing metadata without a body is more correct but opens a new class of inconsistency for materially more work, and is deliberately not in scope.

Receiving side: an inbound CRDT writeback whose resulting body exceeds the note-class thresholds degrades to large-file class on the receiver and is not seeded.

## Acceptance criteria

- [ ] A large-file-class file enqueues no sync item
- [ ] The file does not appear in another device's sidebar
- [ ] The CRDT writeback suite asserts that an inbound body over the thresholds degrades to large-file class and is not seeded
- [ ] Receiving a stream of incremental CRDT updates that add up to an oversized body does not freeze the receiver
- [ ] Note-class files sync exactly as they do today

## Blocked by

- #1459

h4yfans:
Shipped in #1476 (squash `b3a38420`).

**Sending side.** A large-file-class file now enqueues no sync item at all, journal entries included, so it never draws a row on another device that could not be opened. `SyncNoteCreateOptions.initCrdt` became `sizeClass`, since "no CRDT doc" and "no sync item" are one decision.

**Receiving side.** `performWriteback` classifies the body it serialized. The file still lands on disk — it is the user's data — but the renderer is told the note degraded (`sizeClass: 'large-file'`, `contentOmitted: true`) instead of being handed the body, which `note.tsx` would feed straight into the editor. `replaceNoteBodyInCrdt` also refuses large-file-class markdown, which closes the loop where a receiver writes the oversized file out and the watcher feeds it right back in.

No protocol, schema or contract change: an older peer keeps sending oversized notes and a new receiver keeps every byte, writes the file, and opens it read-only instead of freezing.

Known limitation, out of scope here: a file classified large-file at ingest stays unsynced even if it is later edited down under the limits.

PR #1479, opened 2026-08-15
perf(ingest): a sidebar row from stat alone, metadata on idle

### LINKED ISSUE #1460: Sidebar row for a new vault file without reading the file
## Parent

- #1468

## What to build

Everything the sidebar needs to show a row for a new vault file is the path, the title and the filesystem timestamps — a `stat` and a filename. Today the markdown-add handler reads the whole file into a single JS string, hashes it character by character, extracts metadata with several full passes, puts the full parsed content on a projection event and inserts it into FTS, all synchronously on the main process, before the row is emitted.

At 250 MB that is a multi-hundred-megabyte allocation and a major GC pause even after the CRDT seed is gone (#1459).

Split ingest into tiers:

- **On add, O(1)** — `stat` plus path and title produce the sidebar row. The file is not read. Applies to both file classes.
- **On idle, smallest file first** — metadata and FTS backfill, off the main thread, chunked for large files. This fills in word count, snippet and search.

Word count and snippet become nullable on the note list item and are populated by the backfill, so this changes a renderer-facing contract.

Note that V8 caps a single string at 536,870,888 characters, so reading a file above ~512 MB as one string throws. Tier 0 never reads, so it is unaffected; the backfill must respect that ceiling and chunk or skip rather than throw.

## Acceptance criteria

- [ ] A 250 MB markdown file pasted into the vault produces a sidebar row with no main-process stall
- [ ] No full-file read, content hash or FTS insert happens on the add path
- [ ] Word count and snippet are nullable on the note list item, and the IPC contract is updated
- [ ] `pnpm ipc:generate` then `pnpm ipc:check` green
- [ ] The idle backfill populates word count, snippet and search for both file classes, smallest first
- [ ] The backfill does not throw on files above the V8 single-string ceiling
- [ ] Existing vaults open with no reindex and no FTS rebuild
- [ ] Renderer handles a row whose word count and snippet are not yet populated

## Blocked by

- #1459

h4yfans:
Closes #1460. Part of EPIC #1468. Builds on #1474.

## What was still on the add path

The CRDT seed is gone from ingest (#1459), but `handleMarkdownFileAdd` still did all of this before emitting the sidebar row:

1. `safeRead` — the whole file into one JS string
2. `parseNote` / gray-matter over it
3. `generateContentHash`, char by char
4. the full parsed body onto a projection event
5. `flushProjectionEvents` → FTS insert

At 250 MB that is a multi-hundred-megabyte allocation and a major GC pause, to draw one row. Everything the row actually needs is a `stat` and a filename.

## Tiers

**T0, on `add`, O(1).** `stat` + path + title produce the row. The file is not read, for either file class. `wordCount` is published as `null` — unknown, not zero.

**T1, on idle, smallest file first.** `vault/ingest-backfill.ts`. Metadata, tags, links, search, and the CRDT decision, which has to live here: note class needs the *largest block* measured, and the add path never reads the file.

- Note class (≤ `NOTE_MAX_BYTES`) is cheap to read whole and takes the ordinary parse-and-sync path — identical output to before.
- Large-file class is streamed through `vault/file-scan.ts` in 64 KB chunks. Nothing is ever held as one string, so a file past V8's 536,870,888-character ceiling is measured instead of throwing `ERR_STRING_TOO_LONG`. The first 256 KB reach search and the snippet; the counts and the hash describe the whole file.

Tags, properties and links stay empty for large-file class on purpose — a log dump's `#hashtags` are not the user's vault structure, and the file is read-only anyway.

## Keeping rename detection

Rename detection is the one thing on the add path that still needs a content hash: chokidar reports a Finder rename as `unlink` then `add`, and the hash ties them together. Two changes:

- **The hash is computed only while a delete is in flight** (`hasPendingDeletes()`), so an ordinary paste pays nothing — and it is computed by *streaming*. `createContentHasher` folds djb2 chunk by chunk to exactly the value `generateContentHash` returns for the whole string (asserted in `file-scan.test.ts`). A file too big to read is now hashable where it previously was not.
- **A pending delete whose own row was never backfilled has no cached hash to match against.** Renaming a just-pasted file would have read as a delete plus a new note, losing its id. Those rows also register under size + mtime, which a rename leaves untouched.

## Contract

`NoteListItem.wordCount` becomes `number | null` and `snippet` becomes `string | null` in `packages/contracts`, `packages/rpc`, `vault/notes-crud.ts` and the watcher's local copy. Null means "not measured yet". `pnpm ipc:generate` then `pnpm ipc:check` are green; no renderer surface dereferences the field (folder views use `NoteWithProperties`, a separate contract, unchanged).

## Backward compatibility

No migration, no reindex, no FTS rebuild, no `FileType` change. The cache columns were already nullable — `persistFileNote` has always written nulls into them.

The one hazard is a stat-only projection landing on a row that already has real measurements: it updates identity only and never writes nulls over them. Covered by *"keeps the measurements a rename cannot have changed"*.

## Mutation checks

Every guard was reverted and the suite re-run; each went red on the test that owns it.

| # | reverted | test that went red |
|---|---|---|
| M1 | `hasPendingDeletes()` guard | lists a newly added file without reading it |
| M2 | tier-0 nulls → zeros | lists a newly added file… (+2) |
| M3 | `bodyUnread` update guard | keeps the measurements a rename cannot have changed |
| M4 | search-projector null skip | leaves a note out of the index until its body has been read |
| M5 | streaming hash → head only | hashes a multi-chunk file to the same value as hashing it whole |
| M6 | smallest-first ordering | backfills the smallest file first |
| M7 | stat-key rename fallback | detects a rename of a file whose body was never read (+1) |
| M8 | bounded head | keeps only the requested head in memory |
| M9 | word count across chunks | counts words and characters across chunk boundaries (+2) |
| M10 | backfill CRDT seed | still initiates a CRDT seed for a well-formed note |

## Verification

`pnpm lint` · `pnpm typecheck --force` · `typecheck:test` · `check:architecture` · `check:contracts` · `i18n:check` · `ipc:generate` + `ipc:check` · `git diff --check` — all exit 0.

Suites run per project from `apps/desktop`: shared 2497 · main 6589 · main-integration 9 · preload 30 · renderer 7058, all passing.

`pnpm docs:impact --strict` green after the docs update; `pnpm docs:build` green.

### DIFF
--- a/apps/desktop/src/main/projections/projectors/embedding-projector.ts
+++ b/apps/desktop/src/main/projections/projectors/embedding-projector.ts
@@ -225,6 +225,13 @@ export function createEmbeddingProjector(
         return
       }
 
+      // Body not read yet (tier 0 of ingest); the idle backfill republishes the
+      // note with its body. Embedding the title alone would be worse than
+      // waiting, and it would drop any embedding the note already has.
+      if (note.parsedContent === null) {
+        return
+      }
+
       // During the initial index pass, embedding is deferred: indexVault awaits
       // this projector per file (before the vault is marked open), and loading the
       // ~23MB model + running CPU inference here is what stranded vault-open for
--- a/apps/desktop/src/main/projections/projectors/note-derived-state-projector.ts
+++ b/apps/desktop/src/main/projections/projectors/note-derived-state-projector.ts
@@ -32,19 +32,29 @@ const RECONCILE_STAT_CONCURRENCY = 8
 function persistMarkdownNote(note: Extract<NoteProjectionRecord, { kind: 'markdown' }>): void {
   const db = getIndexDatabase()
   const existing = getNoteCacheById(db, note.noteId)
+  // Tier 0 of ingest publishes identity only. Its null measurements mean
+  // "not read yet", so they must not be written over a row that already
+  // carries real ones, and no body-derived state can be rebuilt from them.
+  const bodyUnread = note.parsedContent === null
 
   if (existing) {
-    updateNoteCache(db, note.noteId, {
-      path: note.path,
-      title: note.title,
-      emoji: note.emoji,
-      localOnly: note.localOnly,
-      contentHash: note.contentHash,
-      wordCount: note.wordCount,
-      characterCount: note.characterCount,
-      snippet: note.snippet,
-      modifiedAt: note.modifiedAt
-    })
+    updateNoteCache(
+      db,
+      note.noteId,
+      bodyUnread
+        ? { path: note.path, title: note.title, modifiedAt: note.modifiedAt }
+        : {
+            path: note.path,
+            title: note.title,
+            emoji: note.emoji,
+            localOnly: note.localOnly,
+            contentHash: note.contentHash,
+            wordCount: note.wordCount,
+            characterCount: note.characterCount,
+            snippet: note.snippet,
+            modifiedAt: note.modifiedAt
+          }
+    )
   } else {
     insertNoteCache(db, {
       id: note.noteId,
@@ -53,6 +63,7 @@ function persistMarkdownNote(note: Extract<NoteProjectionRecord, { kind: 'markdo
       emoji: note.emoji,
       localOnly: note.localOnly,
       fileType: 'markdown',
+      fileSize: note.fileSize ?? null,
       contentHash: note.contentHash,
       wordCount: note.wordCount,
       characterCount: note.characterCount,
@@ -62,6 +73,8 @@ function persistMarkdownNote(note: Extract<NoteProjectionRecord, { kind: 'markdo
       modifiedAt: note.modifiedAt
     })
   }
+
+  if (bodyUnread) return
 
   setNoteTags(db, note.noteId, note.tags)
   setNoteProperties(db, note.noteId, note.properties, (name, value) =>
--- a/apps/desktop/src/main/projections/projectors/search-projector.test.ts
+++ b/apps/desktop/src/main/projections/projectors/search-projector.test.ts
@@ -274,6 +274,54 @@ describe('search projector', () => {
     ).toHaveLength(0)
   })
 
+  it('leaves a note out of the index until its body has been read', async () => {
+    // #given the projection vault ingest publishes from `stat` alone: identity
+    // only, every body-derived field null
+    const projector = createSearchProjector(() => vaultDir)
+    const statOnlyEvent = {
+      type: 'note.upserted',
+      note: {
+        kind: 'markdown',
+        noteId: 'note-1',
+        path: 'notes/pasted.md',
+        title: 'Pasted',
+        fileType: 'markdown',
+        localOnly: false,
+        contentHash: null,
+        wordCount: null,
+        characterCount: null,
+        snippet: null,
+        date: null,
+        emoji: null,
+        createdAt: '2026-01-01T00:00:00.000Z',
+        modifiedAt: '2026-01-01T00:00:00.000Z',
+        parsedContent: null,
+        tags: [],
+        properties: {},
+        wikiLinks: []
+      }
+    } as const
+
+    // #when
+    await projector.project(statOnlyEvent)
+
+    // #then no FTS write happens on the add path at all
+    expect(getFtsCount(indexDb.db as never)).toBe(0)
+
+    // #when the backfill republishes the same note with its body
+    await projector.project({
+      ...statOnlyEvent,
+      note: { ...statOnlyEvent.note, parsedContent: 'measured at last', wordCount: 3 }
+    })
+
+    // #then it becomes searchable
+    expect(
+      indexDb.db.all<{ id: string }>(
+        sql`SELECT id FROM fts_notes WHERE fts_notes MATCH ${'measured'}`
+      )
+    ).toHaveLength(1)
+  })
+
   it('reconcile sweeps duplicate rows left behind by earlier versions', async () => {
     seedMarkdownNote('note-1', 'notes/searchable.md', 'Disk content', ['alpha'])
     seedTask('task-1')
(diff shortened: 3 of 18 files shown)


h4yfans:
Self-review found and fixed one real bug in the first push.

`syncNoteStatToCache` went through `saveCanonicalNote` for both callers. That upsert writes every column it is given, and `buildCanonicalNoteMetadata` materialises `propertyDefinitionNames` as `null` when no `properties` are supplied — which a stat-only write never has. So a Finder rename erased the note's `propertyDefinitionNames` until the re-queued backfill restored it, or permanently if the app quit inside that window. That is the same class of bug the `attachmentReferences` comment in `domain-notes/commands.ts` was written for.

Fixed by splitting the canonical write on `{ isNew }`: a new file gets the full upsert, a rename gets an identity-only `updateNoteMetadata` (path, title, journalDate, modifiedAt). Covered by *"keeps the measurements a rename cannot have changed"*, which asserts before the re-queued backfill runs so the window is actually under test. Mutation-checked: forcing the `isNew` branch always-true turns that test red.

Full verification re-run after the fix — lint, typecheck (forced), typecheck:test, architecture, contracts, i18n, ipc:generate + ipc:check, git diff --check all exit 0; main 6589, main-integration 9, preload 30, renderer 7058 passing; docs:impact strict green.

h4yfans:
Second self-review fix, pushed.

`scheduleDrain` returned early when a drain was already in flight. A file enqueued in the window between `runQueue` finding the queue empty and `drainIngestBackfill` clearing `inFlight` would see a run in flight, skip arming the timer, and then never be scheduled — it would sit unmeasured until some *other* file was added or that file changed on disk.

Removed the `inFlight` half of the condition rather than adding a re-check after the run: arming the timer during a run costs nothing, because the callback goes through `drainIngestBackfill`, which joins the run in flight instead of starting a second one. Race-free by construction rather than by patch.

Re-verified: lint, typecheck (forced), typecheck:test, architecture, contracts, i18n, ipc:check, git diff --check all exit 0. Suites: shared 2497 · main 6589 · main-integration 9 · preload 30 · renderer 7058, all passing.

h4yfans:
Rebased onto main after #1476 (#1461) and #1478 (#1463) landed.

**#1476 changed the shape of the decision this PR defers.** `SyncNoteCreateOptions.initCrdt` is gone; `syncNoteCreate` now takes `sizeClass` and returns before `enqueueLocalSyncCreate` for large-file class, so that class gets neither a CRDT doc nor a sync item. I took main's version of that decision rather than reinstating `initCrdt`, and it removed work rather than adding it:

- My `initNoteCrdt` split in `notes/runtime-effects.ts` is no longer needed — `syncNoteCreate` already gates both effects on the class. That file is now **byte-identical to main**; this PR no longer touches it.
- Tier 0 now calls **neither** `syncNoteCreate` nor `enqueueJournalCreate`. Both are gated on a class the add path cannot know, because the class needs the largest block and the add path never reads the file. The backfill makes both calls once it has measured. That is strictly better under #1461's rule than what I had: a large file is never even transiently given a sync item that would have to be withdrawn.
- Main's two new journal tests (`enqueues no journal sync item for a large-file-class journal entry`, `still enqueues a note-class journal entry for sync`) now resolve at tier 1, so they drain the backfill. Both still assert exactly what #1461 wrote them to assert.

**#1478** lowered `NOTE_MAX_BYTES` to 1 MB. No fixture here assumed 2 MB — the over-ceiling test writes 2.7 MB and the block-bound test relies on shape, not size — so nothing needed changing.

Mutation checks re-run against the rebased tree, including two new ones for the effects that moved:

| # | reverted | test that went red |
|---|---|---|
| M1 | `hasPendingDeletes()` guard | lists a newly added file without reading it |
| M3 | `bodyUnread` update guard | keeps the measurements a rename cannot have changed |
| M10 | backfill's `syncNoteCreate` call | lists a pasted log dump… + still initiates a CRDT seed… |
| M11 | `isNew` canonical split | keeps the measurements a rename cannot have changed |
| M12 | backfill hardcodes `sizeClass: 'note'` | lists a pasted log dump but does not initiate a CRDT seed for it |
| M13 | journal large-file gate | enqueues no journal sync item for a large-file-class journal entry |

Full verification re-run after the rebase: lint, typecheck (forced), typecheck:test, architecture, contracts, i18n, ipc:generate + ipc:check, git diff --check all exit 0. Suites: shared 2505 · main 6598 · main-integration 9 · preload 30 · renderer 7058, all passing. docs:impact strict and docs:build green.

h4yfans:
Rebased again — #1472 landed while I was verifying, and it adds a rule this PR has to carry.

#1472 split the tag array handed to the CRDT from the one written to the index: `syncResult.tags` merges the body's `#hashtag`s in for search, the tag hub and the graph, but the CRDT's tag array is what write-back serializes into the file's `tags:` block — so seeding merged tags injects a `tags:` block into a note that never had one (#1454). The seed must get `extractTags(parsed.frontmatter)` only.

Since this PR moves the seed from the add path to the backfill, that rule moved with it: `ingest-backfill.ts` now computes `declaredTags` and hands those to `syncNoteCreate` and `initializeJournalCrdt`, while the index still gets the merged set through `syncNoteToCache`. #1472's two tests (`seeds the CRDT with the declared tags only…`, `hands the CRDT no tags at all…`) now drain the backfill and assert exactly what they asserted before.

Mutation-checked as M14: replacing `declaredTags` with the merged `tags` turns both of those tests red.

Full verification re-run on the rebased tree: lint, typecheck (forced), typecheck:test, architecture, contracts, i18n, ipc:generate + ipc:check, git diff --check all exit 0. Suites all green with clean exit codes this time — shared 2505 · main 6602 · main-integration 9 · preload 30 · renderer 7063. docs:impact strict green.

Issue #1464, opened 2026-08-15
In-file incremental search in the large-file viewer


h4yfans:
## Parent

- #1468

## What to build

Large-file-class files never enter FTS, so the global search box will not find anything inside them. The viewer from #1462 needs its own find-in-file.

Incremental search over the chunked reader: match as the user types, report hit counts, jump between hits, and highlight matches in the visible window without materializing the whole file. Search runs off the main thread and stays responsive at 2 GB.

## Acceptance criteria

- [ ] Find-in-file works inside the large-file viewer at sizes up to 2 GB
- [ ] Searching does not materialize the file as a single string
- [ ] Next/previous hit navigation, with a hit count
- [ ] Matches highlighted in the visible window
- [ ] Search runs off the main thread; typing stays responsive
- [ ] Keyboard shortcut matches the app's existing find affordance

## Blocked by

- #1462

h4yfans:
Shipped in #1482 (`bb2ba2154`).

`Cmd/Ctrl+F` in the large-file viewer now opens the same find bar the rest of the app uses, over a streaming pass in the main process — the file is never held, never decoded whole, and never materialized as a string. `useFindInPage` is switched off on that surface, and Edit > Find and the note menu route to the viewer, so one keystroke means one bar.

Every acceptance criterion, and where it is proven:

- **Works up to 2 GB** — measured, not extrapolated. The env-gated stress test writes a real 2 GB fixture to a temp dir and searches it: scan 542 ms, mid-file seek 0 ms, **search 6913 ms**, 33 554 432 matches counted. No fixture in the repo or CI.
- **Does not materialize the file** — matching is over encoded bytes through fixed windows; a test drives the same code paths with a reader that refuses any window over 64 bytes.
- **Next/previous with a hit count** — `Enter` / `Shift+Enter`, wrapping both ways.
- **Matches highlighted in the visible window** — drawn from line text already on screen, re-found with the same ASCII fold and the same non-overlapping advance the pass uses, so the Nth match here is the pass's ordinal N.
- **Off the main thread, typing stays responsive** — the existing large-file worker entry now carries both whole-file passes; 150 ms debounce, and a superseded query cancels the thread rather than letting it finish crossing the file.
- **Shortcut matches the existing affordance** — the same capture-phase `Cmd`/`Ctrl+F` binding `use-find-in-page.ts` uses.

[...]

PR #1480, opened 2026-08-15
feat(notes): read a large vault file instead of only explaining it

### LINKED ISSUE #1462: Streaming read-only viewer for large vault files
## Parent

- #1468

## What to build

Replace the placeholder from #1459 with a real read-only viewer, so a large log or text file dropped in the vault can actually be read.

Clicking a large-file-class file opens a virtualized read-only view that scrolls smoothly at sizes up to 2 GB, carrying a visible "read-only, not synced" badge so it never fails silently. No editing, no BlockNote, no CRDT.

The file must never be materialized as a single string. V8 caps a string at 536,870,888 characters, so a 512 MB file cannot be held whole either — the viewer reads windows through a file handle at byte offsets and builds a line-offset index in one streaming scan off the main thread. Once chunked reading exists, 512 MB and 2 GB cost the same, which is why the ceiling is 2 GB rather than the engine wall.

Above 2 GB the sidebar row still appears; clicking explains why the file cannot be opened rather than reporting a bare failure.

Prior art for virtualization in this repo: the virtualized notes tree and the virtualized task list.

## Acceptance criteria

- [ ] A 2 GB file opens and scrolls without a main-process stall
- [ ] No code path materializes the file as a single string
- [ ] The line-offset scan runs off the main thread and reports progress
- [ ] The "read-only, not synced" badge is present and legible
- [ ] Above 2 GB the row still appears and clicking gives an explanatory message
- [ ] Reduced-motion and RTL respected; new markup uses logical Tailwind properties

## Blocked by

- #1459

h4yfans:
Closes #1462. Part of #1468.

#1459 stopped an oversized markdown file from becoming a CRDT note and put a panel where the editor used to be. The panel says why the file will not open; it does not let anyone read the file. This replaces it with a read-only viewer that scrolls files up to 2 GB.

## The shape of it

**The file is never a string.** V8 caps one at 536,870,888 chars, so 512 MB is already out of reach and 2 GB is not close. Every read is a window through a file handle at a byte offset (`large-file-index.ts`), and the only whole-file pass is a newline scan that keeps a **sparse** table of line starts — doubling its stride whenever the table fills, so the table is 512 KB for a 2 GB file as much as for a 2 MB one. Reading line L is a seek to checkpoint `floor(L / stride)` plus a short forward walk.

**The scan runs off the main thread.** `large-file-index-worker.ts` is a `worker_threads` entry, registered in `electron.vite.config.ts` and driven by `large-file-index-bridge.ts`, which reports progress as it goes. If the thread cannot start — a build shipped without the entry, a platform that refuses the spawn — the same scan runs in-process through async windows, which yields between chunks and so cannot lock the process the way the BlockNote parse did. A viewer that opens on the wrong thread beats a viewer that does not open. A test asserts the rollup entry exists, because losing it degrades silently.

**Above 2 GB** — SQLite's `MAX_LENGTH`, and where that number comes from — the row stays in the sidebar and opening it names the size, names the limit, and offers Open in default app / Show in folder. A bare "failed to open" was the thing to avoid.

**Read-only, visibly.** The badge `Read-only · not synced` sits in a header bar for as long as the file is open, next to the reason it is not editable. There is no editing surface at all: no BlockNote, no Y.Doc, nothing that could write back.

Sessions live in the main process (`large-file-session.ts`): one open handle plus its index, reused across tab switches when size and mtime are unchanged, rescanned when they are not, and evicted oldest-first at 4 so the app never accumulates handles. A read against a session that is gone returns `null` and the renderer reopens rather than erroring.

Lines longer than 64 KB (minified JSON) come back cut and marked `long line cut here` — returning them whole would put the file back in one string.

## How the 2 GB claim was covered

`large-file-index.stress.test.ts` is env-gated (`MEMRY_LARGE_FILE_STRESS=1`), never runs in CI, and writes its fixture to a temp dir it then deletes. **I ran it**, on a real 2 GB file of 33,554,432 lines:

```
2 GB: scan 635 ms, mid-file seek 0 ms
```

with exact assertions on the line count, on the checkpoint table staying inside its cap, and on progress reaching the file size.

What I executed vs. reasoned about, precisely:

- **Executed:** the 2 GB scan, the 2 GB mid-file seek, and reads of the last lines of a 2 GB file, on this machine, at the numbers above. The over-ceiling path against a real `stat` on a real sparse file of `VIEWER_MAX_BYTES + 1` bytes, and the boundary itself at exactly `VIEWER_MAX_BYTES`. Every unit test below.
- **Reasoned about, not executed:** the end-to-end Electron path — a 2 GB file dropped into a real vault, opened from a real sidebar row. The renderer half is covered by component tests against mocked IPC, and the main half by tests against real files on disk, but the two have not been run together in a packaged app. The hand test is the thing that closes that gap.

The first fixture I wrote *was* sparse (`truncate` to 2 GB), and it was wrong: NUL bytes hold no newlines, so it was one 2 GB line and exercised only the truncation path. The committed fixture writes real 64-byte lines.

## Tests

19 mutation checks, one per guard — production change reverted, test confirmed red, restored. **All 19 went red.** Line count / trailing newline · checkpoint stride doubling · per-line byte cap · CRLF strip · fragment copying off the reused read buffer · progress reporting · the 2 GB ceiling · session reuse · rescan on change · handle eviction · scan-failure reporting · the in-process fallback · the worker rollup entry · the read-only badge · the over-ceiling explanation · handle release on unmount · the real progress value · the cut-line marker · the note-page large-file gate.

The `readFile`-is-impossible test is the one that carries "no code path materializes the file as a single string": it breaks `fs.readFile` outright and asserts that opening and reading still work.

## Verification

`pnpm lint` (0 errors) · `pnpm typecheck` (17/17) · `pnpm --filter @memry/desktop typecheck:test` · `pnpm check:architecture` · `pnpm check:contracts` · `pnpm --filter @memry/desktop i18n:check` · `pnpm ipc:generate` + `pnpm ipc:check` · `git diff --check` · `pnpm docs:impact --strict` + `pnpm docs:build`.

Suites, run per project from `apps/desktop`: shared 2497 · main 6609 · main-integration 9 · preload 30 · renderer 7067. All green, before and after the rebase onto main.

## Compatibility

Additive only. Three new invoke channels and one event on `NotesChannels`; no schema change, no migration, no reindex, no FTS rebuild. `FileType` is untouched. A note without `sizeClass` reads as note class exactly as before, so existing vaults open unchanged.

## Seam for #1464

In-file search runs over the session that is already open — same handle, same index — rather than reopening the file. `withSession(sessionId, (read, index) => …)` in `large-file-session.ts` is that seam, and the chunked reader (`readLines`) and the index (`scanLineIndex`, `LineIndex`) are module-level exports in `large-file-index.ts`, not sealed inside the React component. No search is implemented here.

## What to look at by hand

Drop a ~50 MB file into the vault and open it: progress bar, then lines, with the badge above them. Scroll to the end. Then try something over 2 GB — the row should still be there and the click should explain itself.


h4yfans:
Pushed a fix for the three CodeQL alerts. All three were real; none are suppressed.

**`js/file-system-race` — `large-file-session.ts`.** Was `fs.stat(path)` then `fs.open(path)`: the file measured was not guaranteed to be the file opened, and the whole session rests on that measurement. Now the handle is opened first and everything — the size check, the ceiling test, the mtime used for cache invalidation — is measured through `handle.stat()`. The path is resolved exactly once per session. The reuse and too-large paths close the handle before returning.

**`js/insecure-temporary-file` — `large-file-index-bridge.ts`.** The alert traced a predictable `os.tmpdir()` path from a test into the bridge's own `fs.open`. The bridge no longer opens anything: `buildLineIndex` now takes the caller's `ByteReader`, and the in-process fallback scans through the session's already-open handle instead of re-resolving the path. The worker still takes a path, because a thread cannot inherit a file handle — that is the only remaining resolve, and it is on the other side of a thread boundary. This is also the better design: one handle per session, whichever path the scan takes.

**`js/file-system-race` — `large-file-index.stress.test.ts`.** Open-for-write, close, reopen-for-read is now a single `'w+'` handle.

## On the renderer-controlled inputs

Already covered, and now with tests naming it:

- **Paths.** The renderer cannot send one. `notes:large-file-open` takes a *note id*; the path comes from the index DB via `getNoteCacheById` and `toAbsolutePath`. That was already true, but `toAbsolutePath` is `path.join(vaultRoot, relative)`, which resolves `../` happily — so `openLargeFileSession` now checks containment against `getVaultRoot()` before anything is opened, and returns `missing` otherwise. Test: *refuses a cache path that resolves outside the vault*.
- **Byte offsets.** The renderer never sends one. `notes:large-file-read-lines` takes `{ sessionId, startLine, count }`; every byte offset comes from the line index built by the main process. `sessionId` is an opaque UUID and an unknown one returns `null`.
- **Bounds.** `LargeFileReadLinesSchema` caps `count` at 1..2000 and requires `startLine >= 0`; `readLines` returns an empty window for any `startLine` at or past `lineCount`, off the index alone, reading nothing. Tests: *returns nothing for a start line past the end, without reading a byte* (a reader that throws if called) and *clamps a line window past the end of the file* at the session level.

## Mutation checks on the new guards

- vault-root containment → **RED**
- no I/O for a line past the end → **RED**
- fallback reads the caller's handle → **RED**
- measure through the open handle → **stays GREEN, and I am not going to pretend otherwise.** `fs.stat(path)` and `handle.stat()` agree whenever nothing races, and a deterministic unit test can only produce that race by hooking `fs`, which would test the hook. CodeQL is the check that this one holds; the behavioural tests around it (ceiling, reuse, rescan-on-change) all still pass through the new code path.

Full verification re-run after the fix, not just the changed tests: lint (0 errors) · typecheck 17/17 · `typecheck:test` · architecture · contracts · i18n · `ipc:check` · `git diff --check` · docs impact + build. Suites: shared 2497 · main 6611 · main-integration 9 · preload 30 · renderer 7067. The 2 GB stress test was re-run after the handle change: scan 528 ms, mid-file seek 0 ms.

h4yfans:
Rebased onto main (now `eb1c6520b`), picking up #1478's 1 MB `NOTE_MAX_BYTES`, #1476, #1477 and #1472. Two conflicts, both resolved by keeping both sides: the "Very Large Files" docs section (#1476's sync paragraph + this PR's "Reading a large file" subsection) and a `@/components/note` import list in `note.tsx` (#1472 added `InlineTagsOrigin`; this PR swapped `LargeFileNotice` for `LargeFileViewer`). `mergeable` is now MERGEABLE. All three CodeQL fixes survived, verified by grep and re-run.

**#1476's sync path.** A note can now reach the renderer already flagged `sizeClass: 'large-file'` from an inbound writeback, with no `getNoteById` round trip and therefore no classifier measurements. The viewer degrades correctly: `reason` and `measuredBytes` are optional, the header drops the reason sentence and keeps the badge, and `notes:large-file-open` gets the size from the main process anyway. Added a test for exactly that shape — *opens a note that arrived over sync already flagged, with no measurements*.

## The security question, explicitly

**Is the path constrained to the vault root before any handle is opened? Yes.** `apps/desktop/src/main/vault/large-file-session.ts`:

- L104 `if (!isInsideVault(absolutePath))` → returns `{ status: 'missing' }`. This is **before** L116, the only `fs.open` in the module.
- L183-187 `isInsideVault` resolves both sides and requires `path.relative(vaultRoot, target)` to be non-empty, not start with `..`, and not be absolute — so `../`, an absolute path, and the root itself are all refused.
- The renderer cannot name a path at all: `notes:large-file-open` takes a **note id**, and the path comes from the index DB via `getNoteCacheById` + `toAbsolutePath`. The containment check is there because `toAbsolutePath` is `path.join(vaultRoot, relative)`, which resolves `../` without complaint — so a wrong or tampered DB row would otherwise be honoured.
- Test: *refuses a cache path that resolves outside the vault* — writes a real file one level above the vault, points a cache row at `../escaped.md`, asserts `missing`, no handle, no scan. Mutation-checked: guard removed → **RED**.

**Are offset and length validated? Yes, and the renderer never supplies an offset.**

- `notes:large-file-read-lines` takes `{ sessionId, startLine, count }` — no byte offsets. Every byte offset comes from the line index the main process built. `LargeFileReadLinesSchema` (`packages/contracts/src/notes-api.ts` L199-203) requires `startLine` to be a non-negative integer and `count` to be 1..2000, rejecting anything else at the IPC boundary before the handler runs.
- `apps/desktop/src/main/vault/large-file-index.ts` L177: `startLine >= index.lineCount` returns an empty window immediately, off the index alone, **reading nothing**. L178 guards an empty table. L180 clamps the checkpoint lookup to `checkpoints.length - 1`, so no out-of-bounds index into the table is possible.
- `lineCount` and every checkpoint come from a scan of *this* file through *this* handle, so no offset can point past its end; and a read that does land at EOF returns 0 bytes and ends the loop.
- The per-line cap (`MAX_LINE_BYTES`, 64 KB) bounds the response even when a single line is the whole file, so `count: 2000` cannot be turned into an unbounded payload.
- Tests: *returns nothing for a start line past the end, without reading a byte* (a reader that throws if called — mutation-checked **RED**) and *clamps a line window past the end of the file* at session level with `startLine: Number.MAX_SAFE_INTEGER`.

**Handles.** One per session, at most `MAX_OPEN_SESSIONS` (4) alive, evicted oldest-first, released on close, on unmount, and on `unregisterNotesHandlers` (vault close). All mutation-checked.

Full verification re-run after the rebase: lint (0 errors) · typecheck 17/17 · `typecheck:test` · architecture · contracts · i18n · `ipc:check` · `git diff --check` · docs impact + build. Suites: shared 2505 · main 6624 · main-integration 9 · preload 30 · renderer 7073.

PR #1481, opened 2026-08-15
test(notes): cover the large-file paths codecov found bare

h4yfans:
Follow-up to #1480, which merged with `codecov/patch` red. Test-only; no production code changes.

## What codecov actually flagged

Measured on the merged commit `c08194c78`, per file:

```
File                       % Stmts  % Branch  % Funcs  % Lines  Uncovered
large-file-index.ts          98.16     95.08      100      100  178,193,246
large-file-session.ts        87.80     76.92    83.33    95.83  80,128-129
large-file-index-bridge.ts   83.87     66.66    72.72    92.30  79,84
large-file-index-worker.ts    0.00      0.00     0.00     0.00  18-51   <-- the hole
```

**It is a genuine hole, not an artefact.** The paths that carry the hand test were already the well-covered ones — the chunked reader and line-offset index at 98%/100%, the vault-root containment and offset validation and the above-2 GB refusal all inside `large-file-session.ts` at 95.8% lines. What was bare is the `worker_threads` entry point: every one of its 34 executable lines.

It *looked* untestable, which is why it slipped: the module runs its scan as a side effect of being loaded by a worker thread. It is not untestable — mocking `worker_threads` with a `parentPort` stub and a `workerData` getter, then re-importing per case, drives it end to end.

The renderer's session hook was the second gap, at 75.5% statements. Its uncovered lines were not incidental: they were the recovery paths.

## What is added

**`large-file-index-worker.test.ts`** — the two messages the bridge unpacks: a `done` carrying a `Float64Array` of checkpoints and the right line count, and an `error` when the file will not open. Drift here degrades silently into the in-process fallback for the life of a build, which only ever shows up as "the app feels slow".

**Five viewer/hook tests**, all for things a user reaches by hand:

- a session evicted behind other large files (the main process keeps 4 open) reopens rather than reporting an error
- a scan that finishes *before* `largeFileOpen` resolves — what a small large-file-class dump does — has its `ready` event held, not dropped; dropping it leaves a progress bar that never moves for a file that is already scanned
- pages far behind the viewport are let go, or a viewer walked through a 2 GB file ends up holding it
- a tab closed mid-open still releases the handle it never saw
- a failed open says so, once, in plain language

## The page-cache test is the one worth reading

My first version scrolled nothing — it re-emitted `ready` and asserted the current window still rendered, and it **passed with the trim removed**. The committed version gives the virtualizer stub a scroll position, walks the viewport 60 pages deep against a 48-page cache, scrolls back to the top, and asserts page 0 is fetched a *second* time. That is the observable form of "it was dropped".

## Mutation checks

Every new test, production guard reverted, confirmed red, restored:

| guard | result |
|---|---|
| worker posts the finished index | RED |
| worker reports its own failure | RED |
| page-cache trim | RED |
| reopen after the session is gone | RED |
| buffer events that beat the session id | RED |
| close a session learned after unmount | RED |
| report a failed open | RED |

## After

```
worker      0%     -> 93.75% stmts, 100%  lines
hook        75.5%  -> 92.2%  stmts, 97.8% lines
main-side   86.1%  -> 92.4%  stmts, 97.6% lines
```

## Verification

`pnpm lint` (0 errors) · `pnpm typecheck` (17/17) · `pnpm --filter @memry/desktop typecheck:test` · `git diff --check`. Suites: main 6626 passed, renderer 7078 passed. Docs impact reports no docs-relevant changes, correctly — this is test-only.

### DIFF
--- a/apps/desktop/src/main/vault/large-file-index-bridge.test.ts
+++ b/apps/desktop/src/main/vault/large-file-index-bridge.test.ts
@@ -8,7 +8,17 @@ import { fileHandleReader, type ByteReader } from './large-file-index'
 const mocks = vi.hoisted(() => ({
   /** null = constructing a Worker throws, standing in for a missing bundle. */
   spawn: null as ((path: string) => EventEmitter) | null,
-  spawned: [] as string[]
+  spawned: [] as string[],
+  warned: [] as unknown[]
+}))
+
+vi.mock('../lib/logger', () => ({
+  createLogger: () => ({
+    debug: vi.fn(),
+    info: vi.fn(),
+    warn: (_message: string, context?: unknown) => mocks.warned.push(context),
+    error: vi.fn()
+  })
 }))
 
 vi.mock('worker_threads', () => ({
@@ -56,6 +66,7 @@ describe('buildLineIndex', () => {
   beforeEach(() => {
     mocks.spawn = null
     mocks.spawned = []
+    mocks.warned = []
   })
 
   it('scans on a worker thread, not on the main thread', async () => {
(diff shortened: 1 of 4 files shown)


h4yfans:
Added the lint-baseline fix to this PR, and rebased onto `3bf40f488`.

**`@typescript-eslint/prefer-promise-reject-errors`, `large-file-index-bridge.ts:82`** — real, and worth more than a cast. The bridge rejected with whatever the worker's `error` event carried. A thread that throws a non-Error hands over a bare value; rejecting with it reaches the log as a string with no stack. That matters more here than usual, because the in-process fallback keeps the viewer working — so the log line is the *only* trace that the worker was abandoned at all, and "the app feels slow on big files" is the bug report it turns into. Non-Errors are now wrapped with the original kept as `cause`.

The test asserts the **logged** value, not the rejection, because the rejection is swallowed by design: the fallback succeeds and the scan returns normally. Mutation-checked — replace the wrap with `value as Error` → **RED**.

Lint baseline: **92 → 91**.

**Prettier on `large-file-viewer.tsx:21:5`** — does not reproduce. Line 21 is `export interface LargeFileViewerProps {`, and the quoted text (`| { error?: (...data: unknown[]) => void }`) does not appear anywhere in the file. `npx eslint --no-cache` on it reports exactly one warning, the react-compiler one below. `pnpm lint` across the repo is now 91, i.e. baseline 90 + the one remaining warning, which accounts for everything. That reads like an `--cache` artefact — the repo has bitten us on that before.

**React Compiler, `large-file-viewer.tsx:50` — yes, it opts the component out.** `useVirtualizer` returns functions the compiler cannot memoize safely, so it skips auto-memoizing `LargeFileViewer`. Worth being precise about what that costs, since you are right that this is where memoization matters:

- The container re-renders on every scroll tick. It is cheap — a header, a status branch, and a `map` over ~64 virtual items.
- **The rows are memoized by hand, not by the compiler.** `LineRow` is a separate `memo()` component that never touches the virtualizer, so it compiles normally, and all five of its props are primitives (`line`, `offset`, `text`, `truncated`, `truncatedLabel`). Scrolling therefore re-renders only the rows whose text or offset actually changed, which is the property that has to hold at 33 million lines.
- Page fetches are guarded independently of render count: `ensureRange` skips pages already cached or already in flight, so a re-render storm cannot turn into an IPC storm.

I would rather keep the hand-written `memo` boundary than restructure to satisfy the compiler; it is the same trade the repo's other virtualized views make.

Verification re-run on the combined branch: lint 0 errors / 91 warnings · typecheck 17/17 · `typecheck:test` · architecture · contracts · i18n · `ipc:check` · `git diff --check`. Suites: main 6639 passed, renderer 7078 passed.

PR #1482, opened 2026-08-15
feat(notes): find inside a large file the way find works everywhere else

### LINKED ISSUE #1464: In-file incremental search in the large-file viewer

h4yfans:
Closes #1464. Last child of #1468.

The viewer from #1462 could show a large file but not find anything in one. Large-file-class files never enter FTS, so the global search box is blind to them, and the app's DOM-walking find would confidently report a count of the few dozen virtualized rows on screen as if it were the file's.

## What this adds

`Cmd/Ctrl+F` in the large-file viewer opens the same `FindBar` the rest of the app uses, over a streaming pass in the main process. `Enter` / `Shift+Enter` move between hits, matches are highlighted on the visible lines, and the current hit scrolls into view.

## Matching the existing find affordance

`useFindInPage` (`apps/desktop/src/renderer/src/hooks/use-find-in-page.ts`) binds `Cmd+F` on macOS / `Ctrl+F` elsewhere via a capture-phase `keydown` on `window`. The viewer binds the same key the same way, and `note.tsx` switches `useFindInPage` **off** for a large-file note (`enabled: isActiveNote && !isLargeFile`) so one keystroke never opens two bars. `Edit > Find` (the `memry:menu-find` window event) and the note menu's **Find** item route to the viewer on that surface through an `openFindRef` handshake — the same pattern `focusAtEndRef` already uses for the editor.

## How the search works

One pass over the file through the reader and windows #1462 already built. No second index, no second worker entry: the existing `large-file-index-worker` now takes a job kind and carries both whole-file passes, because they are the same shape of job.

**Bytes, not text.** The pass matches UTF-8-encoded bytes and never decodes. Decoding each window on its own is what corrupts a multi-byte character straddling a boundary; UTF-8 is self-synchronising, so a valid encoded query can only align with a whole character in valid UTF-8. Case folding is ASCII-only and therefore length-preserving — `toLowerCase()` changes the length of some characters (`İ` becomes two), which would shift every offset the pass counts. The documented trade: `error` finds `ERROR`, `é` does not find `É`.

**Boundary-straddling matches.** The pass carries an overlap one byte shorter than the query between windows, and a `resumeAt` byte offset that survives across windows so the tail of one window cannot re-report what its head already reported. Both are covered: `finds a match that straddles a read window`, `counts a match at a window boundary once, not once per window`, `keeps a multi-byte character whole across a window boundary`, `finds a query that is itself multi-byte across a window boundary`.

**Hits are lines, not byte offsets.** A hit is `{ line, ordinal }` — which line, and which occurrence on it. The renderer re-finds occurrences in line text it already holds, with the same fold and the same non-overlapping advance, so its Nth match is the pass's ordinal N. Sending byte offsets would mean translating them back into character positions, which is exactly where a multi-byte character goes wrong.

## Honesty about the count — the 2 GB question

**Streamed, not bounded.** A 2 GB pass is seconds of work (measured below), so the count arrives in pieces on an event and the bar shows `12 so far…` until the pass finishes. A partial count is never rendered as a final one — there is a test that fails if it is.

The *navigable* list is bounded at 2000 positions, because a 2 GB log dump searched for `e` holds hundreds of millions of matches and keeping them would put the file back in memory in another shape. The **count is exact regardless**, and when the list is capped the bar says `1/2000 of 1000000` instead of promising 2000000 stops it cannot make.

## What was executed vs. reasoned about

The 2 GB claim was **measured, not extrapolated**. The env-gated stress test (`large-file-index.stress.test.ts`, off in CI, writes to a temp dir and deletes it) was extended to search the same 2 GB fixture and run locally:

```
MEMRY_LARGE_FILE_STRESS=1 npx vitest run --config config/vitest.config.ts \
  --project main src/main/vault/large-file-index.stress.test.ts

2 GB: scan 542 ms, mid-file seek 0 ms, search 6913 ms
✓ indexes, reads and searches without ever holding the file 8017ms
```

33 554 432 matches counted across 2 GB, hit list capped at 2000, `limited` true, progress reported in pieces. The 6.9 s is exactly why the count streams rather than appearing at the end. No fixture is committed; every other test drives the same code paths at 4-byte to 64-byte windows, which is a boundary every few bytes instead of every few megabytes.

## Security

The query and the session id come from the renderer; the **path never does** — it comes from the session opened inside the vault, and there is no re-`stat`-then-open anywhere. Literal search only, no regex, so there is no backtracking to bound; the query is capped at 200 characters because it is carried between read windows and is therefore part of the memory bound. A query containing a newline matches nothing by construction: every hit is addressed as a line plus an occurrence on it. Temp dirs in tests use `fs.mkdtemp`.

## Compatibility

Additive only. New IPC channel and event, no schema change, no migration, no reindex, `FileType` untouched. Older note rows and existing sessions are unaffected.

## Verification

```
pnpm lint                                  0 errors (94 pre-existing warnings; 0 on files touched)
pnpm typecheck                             17/17 tasks
pnpm --filter @memry/desktop typecheck:test  clean (no file added to the exclude backlog)
pnpm check:architecture                    passed
pnpm check:contracts                       passed
pnpm ipc:generate && pnpm ipc:check        up to date
pnpm --filter @memry/desktop i18n:check    passed
git diff --check                           clean
pnpm docs:impact --base <merge-base> --strict   docs changed on this branch
pnpm docs:build                            build complete

vitest --project shared             2505 passed
vitest --project main               6654 passed, 1 expected fail, 6 skipped
vitest --project main-integration      9 passed
vitest --project preload              30 passed
vitest --project renderer           7081 passed, 6 skipped
```

One main-project file failed on the first full run and passed on two subsequent full runs of the whole project; it was not one of the files touched here and did not reproduce.

## Mutation checks

Every guard was reverted, the test watched go red, and restored. Search core: overlap carry removed (3 red), `resumeAt` removed (1), chunk fold removed (1), naive `byte | 0x20` fold (7), newline-in-query guard (1), empty-query guard (2), `shouldStop` (1), hit cap (1), carry-newline recount (2), progress (1). Bridge: worker never used (2), no fallback (3), cancel a no-op (2), fallback ignores cancellation (1). Session: unknown-session null (1), previous query not cancelled (1), stale progress unfiltered (1), no broadcast (1), dispose leaves search running (1), cancelled reported as complete (1). Renderer: shortcut unbound (7), partial count as final (1), capped list as total (1), no scroll to hit (1), no highlight (2), `prev` walks forwards (1), close keeps query (1), close leaves bar open (1), progress ignored (1). Routing: DOM find left on for large files (1), find always to the DOM walker (1).

Two assertions were strengthened after a mutation survived them.


h4yfans:
Follow-up commit d73b336f1 after self-review:

- Two find-bar states that had no test: a session the main process no longer has, and a search that fails. Both must settle the bar — `searching` that never ends reads as "still counting" forever. The first versions of both assertions passed **before the search had even started**, because the settled state and the pre-search state look identical; they now assert only after the bar is observably searching, and the failure branch is mutation-proven (reverting the clear turns the test red).
- One behaviourally-equivalent mutant, reported rather than hidden: deleting the `if (!result)` session-gone branch makes the code throw on `null.hits` into the `.catch`, which settles the bar the same way. The test asserts the right external behaviour; no mutation can distinguish those two paths.
- A comment on `close()` recording that an in-flight pass is left to finish and its answer dropped — bounded at one per session, cancelled by the next query or by closing the file.

`MEMRY_DOCS_IMPACT_SKIP=1` was needed to push this one commit: the pre-push hook measures from the previously pushed commit, and this commit is tests plus a comment. The branch-level gate is green — `pnpm docs:impact --base c08194c78 --strict` reports `docs changed on this branch`, and the user-facing docs for the feature landed in 2fc88cf4e.

h4yfans:
Second self-review pass found a real bug, fixed in b3f12eb9d.

A session runs **one** search at a time — a newer query cancels the older one. The hook read every `cancelled` result as "a newer query of mine will answer" and returned without settling. That is right when this bar superseded itself (its `queryRef` has already moved on, so it returns earlier anyway), but wrong when something *else* took the session's search: a second window open on the same file, or the file being closed. In that case the bar counted `N so far…` forever against a pass that was no longer running.

A query this bar did not replace now settles. Mutation-checked: restoring the bare `return` turns the new test red.

Also pushed with `MEMRY_DOCS_IMPACT_SKIP=1` — same reason as the previous commit: the pre-push gate measures from the last pushed commit, and this one is a renderer fix plus its test with no user-facing change beyond what the docs already describe. The branch-level gate stays green against the merge base.

h4yfans:
Rebased onto `4dc501f0b` (#1481) — force-pushed as `ac5326502`. What the resolution kept:

- **`large-file-index-worker.test.ts`** was an add/add conflict: #1481 and this branch both created it. Resolved onto #1481's harness (the `parentPort` stub + `workerData` getter + per-case re-import) with its two cases kept verbatim, plus two search-job cases on the same harness. The index job now passes `kind: 'index'` explicitly, which is what the entry reads.
- **`large-file-index-bridge.ts`** auto-merged: #1481's `asError` wrapping is intact and untouched, alongside this branch's shared worker-filename constant and job kind. #1481's test asserting the *logged* value still passes.
- **The same hazard existed in the new search bridge** and I had missed it: it spawns the same worker entry and hides its failures the same way, so a thrown non-Error would have reached the log as a bare string. It now reuses `asError` rather than restating it — the message drops "line-index" because one entry carries both jobs, which #1481's assertions (`toContain('ENOMEM')`, `cause === 'ENOMEM'`) are indifferent to. Mutation-checked: rejecting the bare value again turns the new test red.
- **`large-file-viewer.test.tsx`**: both test sets kept, one conflict in the helper block resolved by keeping both helpers. 22 `it`s total.

Full verification re-run against the rebased branch, not just the affected tests:

```
pnpm lint (changed files)                  0 errors, 1 pre-existing warning (useVirtualizer, from #1480)
pnpm typecheck                             17/17
pnpm --filter @memry/desktop typecheck:test  clean
pnpm check:architecture / check:contracts  passed
pnpm ipc:generate && pnpm ipc:check        up to date
pnpm --filter @memry/desktop i18n:check    passed
git diff --check                           clean

vitest --project shared             2505 passed
vitest --project main               6669 passed, 1 expected fail, 6 skipped
vitest --project main-integration      9 passed
vitest --project preload              30 passed
vitest --project renderer           7089 passed, 6 skipped
```

On `codecov/patch`: the worker entry point is covered by four cases through the #1481 harness — both jobs' success paths and both jobs' open-failure paths — and the bridge by six, including the fallback, both cancellation paths, and the non-Error log. The search core is covered by 16. I will not merge past a red patch check.

h4yfans:
Merged as `bb2ba2154`. State of main afterwards, reported rather than glossed:

Everything on the PR was green at merge, including `CodeQL` and `codecov/patch`. On main, seven of the eight workflows on the merge commit are green (Static checks, Unit & integration tests, Coverage thresholds, Packaged runtime smoke, Security CI, Sync Server CI, Code Quality, React Doctor, docs + sync-server deploys).

**Desktop CI is red on main, and was red before this merge.** Four E2E jobs fail — `Electron E2E smoke`, `smoke (macOS)`, `smoke (Windows)`, `full (15/16)` — and the failing tests are the same three every time:

```
tasks.e2e.ts:308  T541: should insert at the top when dropping on a target priority header
tasks.e2e.ts:354  T541: should highlight the full target section and insert at the hovered row position
tasks.e2e.ts:409  T541: should keep intermediate sections visually stable during cross-section drags
```

Byte-identical failures, same three tests with all retries exhausted, on `3bf40f488` (#1479) and `4dc501f0b` (#1481) — both before this PR — and unchanged on `bb2ba2154`. 64 other smoke tests pass. E2E is skipped on PRs and only runs on main pushes, which is why this branch could not have caught it either way.

Task drag-and-drop shares no code with the large-file viewer, and this PR touches no task, DnD or priority-section code. Earliest evidence points at `3bf40f488`; `c08194c78`'s own E2E run was cancelled when the next push superseded it, so it cannot be used as the last-known-good point. Worth a separate issue against #1479 — flagging rather than fixing here.

Issue #1557, opened 2026-08-17
`notes.resolveByTitle` has no heading awareness, so CLI and agents miss `[[Note#Heading]]`


h4yfans:
A15 fixed heading-link resolution in the renderer (`resolveWikiLink`): split the target, look the note half up first, fall back to the raw string.

`notes.resolveByTitle` in `packages/app-core/src/notes.ts` was left alone on purpose — its contract is "is there a note with this title", and burying wiki-link grammar in it would make a note genuinely titled `Sprint #4` unresolvable. That is data going invisible, not a missing convenience.

The consequence is that everything resolving titles **outside** the renderer still misses:

- the CLI — `apps/cli/src/run.ts:333`
- the agent MCP surface — `notes.resolveByTitle` is allowlisted in `packages/contracts/src/agent-mcp-channels.ts:26`

An agent asked to follow `[[Meeting#Decisions]]` gets `null`.

## Proposal

Give those callers the same split-first/raw-fallback resolution the renderer uses, built on `splitWikiTarget` from `@memry/shared/wiki-target` — as a layer above `resolveByTitle`, not inside it.

Issue #1556, opened 2026-08-17
Snippets and exports render `[[Note#Heading]]` as `Note#Heading`


h4yfans:
Five places strip wiki links down to plain text and keep the heading half, so a heading link reads as `Note#Heading` in previews, search results and HTML export:

- `packages/app-core/src/markdown.ts:161-162`
- `packages/app-core/src/note-files.ts:189-190`
- `apps/desktop/src/main/vault/journal.ts:415`
- `apps/desktop/src/main/vault/frontmatter.ts:586`
- `apps/desktop/src/main/lib/export-utils.ts:52-54`

Each should render the note half (and the alias, where one is given), matching what the editor chip and the graph already do.

A sixth site, `apps/desktop/src/renderer/src/lib/home/journal-widget-data.ts`, was fixed with A15 because it was worse than cosmetic: it stripped `#` on the next line, welding `Note#Heading` into `NoteHeading`.

Deliberately left out of the A15 fix — these are display, not behaviour, and spreading that diff across six more files would have buried the claim it was making.

Issue #1555, opened 2026-08-17
Junk notes named `Note#Heading` are already in users' vaults — offer a cleanup


h4yfans:
Until the A15 fix, clicking `[[Note#Heading]]` took the `create` branch of `resolveWikiLink` and created a note titled with the whole string. That wrote a real `Note#Heading.md` into the vault and synced it to every device.

The fix stops new ones. It deliberately does **not** touch the existing files: deleting from a user's vault, over sync, is not reversible, and it is their data.

Resolution is safe in the meantime — it looks the note half up **before** the raw string, so `[[Meeting#Decisions]]` reaches `Meeting` even when the junk `Meeting#Decisions` note still exists. The junk is clutter in search, the sidebar and the graph, not a broken link.

## Proposal

A cleanup surface in Settings that:

- lists notes whose title contains `#` **and** whose body is empty or near-empty (the shape this bug produced)
- shows what each one would have resolved to
- deletes only what the user ticks

A note genuinely called `Sprint #4` must be easy to spot and keep, which is why the body check matters and why nothing is preselected.

Related: the A15 fix (`splitWikiTarget` + split-first resolution).

Issue #1547, opened 2026-08-16
Journal editor is standalone Tiptap capped at heading levels 1-3 — migrate it to BlockNote


h4yfans:
## Summary

The journal editor is the only editing surface in the app that is **not** BlockNote. It is a hand-rolled Tiptap editor, and its heading support is explicitly capped at levels 1-3. Journal entries are stored as markdown on disk and sync via CRDT, so any `####`/`#####`/`######` that reaches a journal file from outside the editor is not representable in the editor.

## Evidence

Journal is raw Tiptap, not BlockNote:

- `apps/desktop/src/renderer/src/components/journal/journal-editor.tsx:7-8` — `useEditor` from `@tiptap/react` + `StarterKit`
- `apps/desktop/src/renderer/src/components/journal/journal-editor.tsx:78-80` — `heading: { levels: [1, 2, 3] }`
- `apps/desktop/package.json:202-215` — 15 `@tiptap/*` direct dependencies

Note: `@tiptap` imports also appear under `components/note/**`, but those reach into BlockNote's internal ProseMirror layer (BlockNote is built on Tiptap). Journal is the only surface that instantiates a Tiptap editor of its own.

Editor speaks HTML while storage speaks markdown:

- `journal-editor.tsx:222-223` — `editor.getHTML()` is what is handed to `onContentChange`
- `journal/note-drawer.tsx:157` — `dangerouslySetInnerHTML={{ __html: note.content }}`
- but on disk journals are markdown: `journal/YYYY-MM-DD.md`, `fileType: 'markdown'`
- and they round-trip through CRDT: `apps/desktop/src/main/sync/crdt-writeback.ts:469` `writebackJournal(...)`, `apps/desktop/src/main/sync/journal-sync.ts`

## Why this matters (data fidelity, not just parity)

A journal `.md` can receive heading levels 4-6 from paths that never touch the Tiptap editor:

- an Obsidian vault opened in Memry
- CRDT sync from another device
- the MCP write tools — `vault_create_journal_entry` / `vault_update_journal_entry` take `content_markdown` (`apps/desktop/src/main/agent/mcp/tools/handles-adapter.ts:631`)

With `levels: [1, 2, 3]`, those levels are not representable in the editor.

## Scope

Replace the Tiptap journal editor with BlockNote, reusing the shared `editorSchema` from `packages/editor-schema`, so journal and notes share one editor, one schema and one heading scale.

Known work beyond the component swap:

[...]

PR #1558, opened 2026-08-17
feat(notes): heading links — resolve them, scroll to them, and write them

h4yfans:
> Single PR. #1559 was folded into this branch and closed.

## The bug

`[[Note#Heading]]` is what an Obsidian vault is full of. Memry looked the whole string up as a note title, the lookup missed, and clicking the link took the `create` branch — writing `Note#Heading.md` into the vault and syncing it to every device.

One missed lookup, three faces:

| Surface | Before |
|---|---|
| Click | Creates a junk note named after the link |
| Hover | No preview card at all (`previewByTitle` → null) |
| Backlinks | Target note never lists the link (`setNoteLinks` writes an unresolved row) |

The graph view was the exception — it has always stripped `#` (`app-core/graph.ts:43`). "Backlinks already work" was true only of the graph; notes and the graph were two link pipelines disagreeing.

Reported by Aurelie, 8 Aug. Everyone migrating from Obsidian is affected. And once resolution was fixed, a second half of the report surfaced: there was no way to *write* one of these links either.

## 1 · Resolving them

`splitWikiTarget` (`@memry/shared/wiki-target`), and resolution that looks the **note half up first, falling back to the raw string**. The order carries weight both ways:

- **split-first** reaches `Meeting` instead of the junk `Meeting#Decisions` note this bug already created — without it the fix does nothing for the people it hurt
- **raw fallback** keeps `[[Sprint #4]]` reaching a note really called that; `#` is legal in a filename

The only losing case is a vault holding both `Sprint` and `Sprint #4`, where Obsidian also picks `Sprint`. When nothing matches, the note is created under the **note half** — a heading separator has no business in a filename.

`resolveByTitle` is deliberately untouched: burying wiki-link grammar in "is there a note with this title" would make a note genuinely called `Sprint #4` unresolvable, which is data going invisible. CLI and MCP tracked in #1557.

## 2 · Scrolling to them

The heading rides to the target page as tab view state and is consumed once that editor has produced headings, matched on trimmed, case-folded text — all a link carries. Scroll restore keeps saving but stands down for that mount: the user just named a destination, the fresher of the two intents. It re-arms next visit.

`[[#Heading]]` scrolls in place · `[[A#H1#H2]]` uses the last segment · `[[A#^block-id]]` opens at the top, since block references do not exist here · a heading that never arrives gives up on **the same deadline scroll restore uses**.

Also fixes the outline's `document.querySelector`, which returned whichever pane was first in the DOM — in split view it scrolled the pane the user was not looking at, and a heading link lands on the identical trap. Both now go through one helper that also respects `prefers-reduced-motion`.

## 3 · Writing them

**`#` in the `[[` menu.** Type `[[Meeting#` and the menu lists Meeting's headings. `#` is an ordinary separator, not a magic trigger — headings appear only once the note part is an **exact** title match, so `[[Sprint #` still lists notes and nothing about `[[Sprint #4]]` changes. No "create heading": writing into someone else's note is a different authority.

There is no heading index in either database, so the menu reads the target's body on demand and caches it. That body is markdown while the click side matches BlockNote's plain text, so `## **Q3** review` would have been written in a form that could never land. `extractMarkdownHeadings` strips inline markdown, skips fenced code, drops closing `##`, collapses whitespace. Two details worth review: underscore emphasis is stripped only at word boundaries (the blunt `[*_~]` strip in `app-core` turns `snake_case` into `snakecase`), and setext headings are skipped on purpose (`---` is also a thematic break and a frontmatter fence; a false positive is a pickable heading that can never be found).

**A "Link to note" slash entry** that types `[[` and hands over to the same menu — one search UI, `#` support not written twice.

**Editing a link you already inserted.** This is the case a real user hits: type `[[`, pick from the dropdown, and now the chip is an atom — the caret cannot get inside, a `#` after it is a separate word, one Backspace deletes the whole thing. Backspace or ArrowLeft next to a chip now un-promotes it to `[[Target]]` with the caret at the end of the target, where a `#` belongs. From there it is plain text, so the menu above works unchanged. The caret leaving promotes it back.

Raw text rather than the `hashTag` shrink-the-attribute pattern, which is still the precedent for intercepting a key next to an inline node: that plugin absorbs by character class, and titles and headings are mostly spaces.

Two landmines avoided: the plugin registers through `register-editor-plugin` (a direct `registerPlugin` destroys the Yjs UndoManager — `f0fb0048c`), and `use-editor-sync`'s normalization exempts the caret's block **only** while the caret is inside a raw run. Exempting it unconditionally also stopped a hand-typed `[[Note]]` from becoming a chip until the caret left; without any exemption the run is promoted back on the next keystroke through a whole-document `replaceBlocks` under a typing caret.

⚠️ **Backspace next to a link no longer deletes it outright.** That is a deliberate change for every wiki link, not only heading ones, and it is documented as one.

## Compatibility

No schema change, no migration, no IPC contract change. `target` stays the raw string in the document, the vault file and the shared Y.Doc — this reads it, never rewrites it. Existing junk notes are not touched (deleting from a vault over sync is not reversible); #1555 proposes an opt-in cleanup. Link extraction now indexes heading links under the note, matching the graph; existing vaults keep their rows until each note is next projected — no forced reindex.

## Verification

`typecheck` 17/17 · `lint` 0 errors (98 warnings, unchanged from main) · renderer **636 files / 7418 tests** · shared 145 / 2527 · `i18n:check` · `check:architecture` · `check:contracts` · `docs:impact --strict` · `docs:build` — all green.

**The E2E is the interesting part of this PR's history.** Four runs, four different lessons, none of which the 7400-test renderer suite could have taught:

1. Failed at 2095 → I read it as a real defect and fixed a genuine race (`ContentArea` renders a placeholder until the CRDT binding settles while `onHeadingsChange` fires earlier, so a single lookup misses and never retries). Worth having, but not what the test was hitting.
2. Failed at **exactly 2095 again** — identical to the pixel across two builds, the signature of a measurement scroll cannot affect. It was: subtracting the editor wrapper's rect from the heading's, both inside the scroller.
3. Rewrote it viewport-relative: `scrolled: true, nearTop: false`. Real signal at last — but `toMatchObject` prints only the keys it compared, so the diagnostic fields went unreported.
4. Numbers finally printed: `scrollTop 1950` against `scrollHeight 2813` and `clientHeight 863` — exactly maximum scroll. **The feature had been right all along.** The fixture put the heading at the end of the note, so `scrollIntoView({block: 'start'})` had nothing beneath it to scroll past. Filler now runs on both sides.

E2E jobs are gated on `refs/heads/main`, so these ran via `workflow_dispatch`.

**Both specs now pass** ([run 32040527449](https://github.com/memrynote/memry/actions/runs/32040527449), shard 16):

```
✓ 24 wiki-heading-links.e2e.ts:36 › opens the note and scrolls to the heading instead of creating a note (15.3s)
✓ 25 wiki-heading-links.e2e.ts:159 › still opens a note whose title really contains a hash (15.1s)
```

That run is red overall, but both failing jobs (`E2E full (2/16)`, `E2E smoke (Windows)`) failed at **"Set up job"** — runner provisioning, before any code executes.

## Manually verified

Part 3 has now been exercised in a running app:

- Backspace next to a chip opens it as `[[Target]]` with the brackets dimmed
- typing `#` there opens the heading picker on that note's real headings
- picking one leaves a single chip

Two bugs were found that way and only that way, both from one root — the suggestion menu is a live plugin session, not something derived from text that happens to look like a trigger:

1. Nothing re-opened the menu when editing existing raw text, so `#` did nothing (`11322ffe8`).
2. The menu was then anchored at the end of the target instead of just after the `[[`, so its query began empty and `#` produced the query `#` — no note half, no heading mode, "No notes found" on a note with two headings (`e68e8fcd0`). The final caret position is identical either way, which is exactly why the unit suite stayed green through it; the new test observes the caret from inside the `openMenu` callback, so the order is pinned rather than the outcome.

## Still not covered by E2E

Part 3 has unit tests and a manual pass, but no E2E: the un-promotion, the dimmed brackets, the `#` menu in edit mode. Both bugs above were invisible to 7,400 renderer tests, so a spec for that path is worth adding.

## Follow-ups

#1555 junk-note cleanup · #1556 five remaining snippet/export strippers · #1557 CLI + MCP resolution · #1547 updated with the journal's scope (its click already works; only authoring and in-journal `[[#Heading]]` wait on the BlockNote migration)

### DIFF
--- a/apps/desktop/src/main/vault/frontmatter.test.ts
+++ b/apps/desktop/src/main/vault/frontmatter.test.ts
@@ -163,6 +163,23 @@ describe('frontmatter utilities', () => {
     expect(links).toEqual(['Same Note', 'Other'])
   })
 
+  // A15: the whole `Note#Heading` string used to go in as a title, resolved to
+  // nothing, and the note it named never listed the link as a backlink.
+  it('extractWikiLinks indexes a heading link under its note', () => {
+    const links = extractWikiLinks('See [[Meeting#Decisions]]')
+    expect(links).toEqual(['Meeting'])
+  })
+
+  it('extractWikiLinks collapses heading links onto the note they share', () => {
+    const links = extractWikiLinks('[[Meeting#Decisions]] and [[Meeting#Actions]] and [[Meeting]]')
+    expect(links).toEqual(['Meeting'])
+  })
+
+  it('extractWikiLinks ignores a same-note heading link', () => {
+    const links = extractWikiLinks('Jump to [[#Decisions]] and see [[Other]]')
+    expect(links).toEqual(['Other'])
+  })
+
   it('extractTags trims and preserves case', () => {
     const frontmatter: NoteFrontmatter = {
       id: 'abc123def456',
--- a/apps/desktop/src/main/vault/frontmatter.ts
+++ b/apps/desktop/src/main/vault/frontmatter.ts
@@ -15,6 +15,7 @@ import {
 } from '@memry/app-core/markdown'
 import { generateNoteId, isValidNoteId } from '../lib/id'
 import { isRelationValue } from '@memry/contracts/relation-uri'
+import { splitWikiTarget } from '@memry/shared/wiki-target'
 
 // ============================================================================
 // Types
@@ -214,6 +215,20 @@ export function validateNoteId(id: string): boolean {
  * Extract all wiki-style links from markdown content.
  * Matches [[Link Title]] and [[Link Title|Display Text]] patterns.
  *
+ * `[[Note#Heading]]` yields `Note`: this table answers "which notes does this
+ * note point at", and a heading link points at the note. Before that, the whole
+ * string went in as a title, resolved to nothing, and the note it named never
+ * listed the link as a backlink. The graph view has always read these targets
+ * this way (`wikilinks` in @memry/app-core/graph), so this makes the two link
+ * pipelines agree rather than introducing a second reading.
+ *
+ * The cost is the `Sprint #4` case: a note really called that is indexed here
+ * under `Sprint`. Navigation and hover still reach it — they fall back to the
+ * raw title against the database, which this function has no access to.
+ *
+ * Existing vaults keep their old rows until each note is next projected; no
+ * reindex is forced for a backlink row.
+ *
  * @param content - Markdown content
  * @returns Array of link targets
  */
@@ -223,7 +238,10 @@ export function extractWikiLinks(content: string): string[] {
   let match
 
   while ((match = linkPattern.exec(content)) !== null) {
-    links.add(match[1].trim())
+    const { note, heading } = splitWikiTarget(match[1])
+    // `[[#Heading]]` addresses the note it sits in — a self-link, not an edge.
+    if (heading !== null && !note) continue
+    links.add(heading !== null ? note : match[1].trim())
   }
 
   return Array.from(links)
--- a/apps/desktop/src/renderer/src/assets/base.css
+++ b/apps/desktop/src/renderer/src/assets/base.css
@@ -2647,6 +2647,13 @@ del.bn-inline-content {
   color: rgb(251, 191, 36) !important;
 }
 
+/* A wiki link opened for editing: the `[[` and `]]` are real characters the
+   user can delete, so they are dimmed rather than hidden — the text still reads
+   as a link without the brackets shouting. */
+.wiki-link-bracket {
+  opacity: 0.4;
+}
+
 /* Inline hash tag styles — needs .dark .bn-editor prefix to beat the wildcard rule */
 .dark .bn-editor .inline-hash-tag,
 .dark .bn-container .inline-hash-tag,
--- a/apps/desktop/src/renderer/src/components/note/content-area/ContentArea.test.tsx
+++ b/apps/desktop/src/renderer/src/components/note/content-area/ContentArea.test.tsx
@@ -2,6 +2,7 @@ import { act, fireEvent, screen, waitFor } from '@testing-library/react'
 import { renderWithProviders as render } from '@tests/utils/render'
 import { beforeEach, describe, expect, it, vi } from 'vitest'
 import * as Y from 'yjs'
+import { WIKI_LINK_EDIT_PLUGIN_KEY } from './wiki-link-edit-plugin'
 
 const contentAreaMocks = vi.hoisted(() => ({
   editor: null as any,
@@ -27,6 +28,7 @@ const contentAreaMocks = vi.hoisted(() => ({
   toastError: vi.fn(),
   defaultFileItemClick: vi.fn(),
   openSuggestionMenu: vi.fn(),
+  registerPlugin: vi.fn(),
   createLinkMentionContent: vi.fn(
     (url: string, domain: string, title?: string, favicon?: string) => ({
       type: 'linkMention',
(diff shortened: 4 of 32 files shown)


h4yfans:
E2E via `workflow_dispatch` (the only way to run it before main): https://github.com/memrynote/memry/actions/runs/32029143864

That run covers `e2e-smoke` (ubuntu/macOS/Windows) and `e2e-full` across 16 shards, which is where the new `wiki-heading-links.e2e.ts` gets picked up — it was deliberately not added to the smoke spec lists, since those are a fixed time budget. The parallel `pull_request` run alongside it is lint + typecheck + unit only and says nothing about scroll.

Staying in draft until this comes back green.

h4yfans:
## The first E2E run found a real defect — in the feature, not the test

[Run 32029143864](https://github.com/memrynote/memry/actions/runs/32029143864), shard 16/16: `opens the note and scrolls to the heading instead of creating a note` failed all three attempts (the other 24 specs in that shard passed, so not a flake).

```
Expected: < 120
Received:   2095
```

The title assertion passed, so navigation was correct — the note simply never scrolled.

**Cause.** `ContentArea` holds its render behind a placeholder until the CRDT binding settles (`ContentArea.tsx:1506`), but `onHeadingsChange` fires as soon as the fragment binds (`use-editor-sync.ts:237`). There is a window where the heading is in the page's `headings` state and its `[data-id]` node is not in the document. The anchor looked **once**, missed, and returned — and a note nobody then edits emits no second set of headings to re-run on.

**Fix** (`842196bc3`): `scrollToHeadingWhenReady` waits on animation frames until the block exists, then keeps re-applying until the offset holds still for three frames, since lazy chunks and the editor mount keep growing the page under an early write. Same deadline scroll restore uses; the block id is read through a ref so late headings are picked up without restarting the deadline.

This is precisely the failure jsdom cannot see — the renderer suite was green through all of it. Worth recording as the argument for keeping the claim in E2E.

Re-running: https://github.com/memrynote/memry/actions/runs/32032090676

h4yfans:
## Correction: the test was measuring something that cannot change

The second run failed again at **exactly 2095** — the same value, to the pixel, as the first run, on a different build. A page that genuinely fails to scroll would still vary a little between runs. An identical number is the signature of a measurement that scroll cannot affect.

It was. The assertion did `heading.top - (editor wrapper).top`, and both elements sit inside the scroller and move together, so the difference is the heading's offset **within the content** — the same number scrolled or not. That assertion could not have passed with a working feature either.

So: **the heading jump is still unproven in both directions.** `842196bc3` fixed a real race found by reading the code — `ContentArea` renders a placeholder until the CRDT binding settles while `onHeadingsChange` fires earlier, so a single lookup can miss and never retry — but the test could not have told us whether that race was what the E2E was hitting, and I should not have read the first failure as confirmation of it.

`b25db474e` polls two independent signals instead: the real scroller (walked up from the editor until something can actually scroll) has moved, and the heading's **viewport-relative** top is where a reader would look for it.

Re-running: https://github.com/memrynote/memry/actions/runs/32033378824

The second spec in the file (`[[Sprint #4]]` still opens the note titled that) has passed in both runs, so resolution — the actual reported bug — is confirmed. It is only the scroll half that is unmeasured.

h4yfans:
Consolidated to a single PR: #1559's work is now on this branch (`ab52050d9`, `4f3ece139`) and that PR is closed.

Latest E2E, on the current head with the fixture fixed: https://github.com/memrynote/memry/actions/runs/32040527449

PR #1559, opened 2026-08-17
feat(notes): a wiki link can name a heading while you are writing it

h4yfans:
> **Stacked on #1558.** Base is `fix/wiki-heading-links`, not `main`. Review that one first; this diff is only the authoring half.

## What this adds

PR #1558 made an existing `[[Note#Heading]]` link *work*. It left the other half of Aurelie's report untouched: there is no way to author one. The `[[` autocomplete searches titles, and `#` just fell into the title search and matched nothing.

**1. `#` mode in the `[[` menu.** Type `[[Meeting#` and the menu lists Meeting's headings; pick one and you get a single link token.

`#` is an ordinary separator, not a magic trigger — the menu only switches to headings once the note part is an **exact** title match, so `[[Meet#` still shows notes and backspacing over the `#` returns to the note list. Alias still works (`[[Meeting#Decisions|the outcome]]`). No "create heading" option: writing into another note is a different authority. Block references are not offered.

**2. A "Link to note" slash entry.** It inserts `[[` and hands over to the menu above rather than opening a second dialog — one search UI, `#` support not written twice, and it teaches the `[[` shortcut instead of competing with it.

## The trap this had to avoid

There is no heading index in either database, so the menu reads the target note's body on demand and caches it. That body is **markdown**, but the click side (#1558) matches against BlockNote's **plain text** of the heading block. A heading written `## **Q3** review` gives `**Q3** review` from markdown and `Q3 review` from the editor — so the link would have been written in a form that could never land.

`extractMarkdownHeadings` therefore strips inline markdown (emphasis, code, links, wiki links), skips fenced code blocks, drops closing `##` sequences, and collapses whitespace the way HTML rendering does. Two details worth a look in review:

- **Underscore emphasis is stripped only at word boundaries.** The blunt `[*_~]` strip that `app-core`'s `stripMarkup` uses would turn `snake_case` into `snakecase` and write an unmatchable link.
- **Setext headings (`===`/`---` underlines) are deliberately skipped.** `---` is also a thematic break and a frontmatter fence; a false positive is a heading you can pick that the click side can never find, whereas a false negative is just a missing row.

## Notes on the implementation

- **Body read goes through `notes:get-by-path`, not `notes:get`** — the latter emits a `note_opened` telemetry event (`notes-handlers.ts:215`), and typing `#` in a link is not opening a note.
- **Two empty states**, not one: "this note has no headings" and "no headings match your filter" would otherwise both read as a lie in one of the two cases.
- No contracts/preload change, so no `ipc:generate`. No schema change; `target` stays one raw string, exactly as on disk.

## Verification

Re-run after rebasing onto #1558's current head (`842196bc3`), not inherited from the pre-rebase run:

```
pnpm typecheck                          → Tasks: 17 successful, 17 total
pnpm lint                               → 0 errors, 98 warnings (same count as the base — this adds none)
                                        → Test Files  635 passed (635)
                                               Tests  7397 passed | 7 skipped (7404)
pnpm --filter @memry/desktop i18n:check → ok: i18n check passed
pnpm check:architecture                 → architecture boundary check passed
pnpm check:contracts                    → contracts boundary check passed
git diff --check                        → clean
```

## Not verified

**The slash-menu handover has not been exercised in a running Electron app.** It goes through BlockNote's own `openSuggestionMenu('[[', { deleteTriggerCharacter: true })`, which inserts the trigger inside the same transaction that sets the plugin meta, so the plugin state cannot be missed — but that is reading the library's source, not clicking the item. A unit test asserts the exact call. Worth a manual pass before this leaves draft.

No E2E was added here. Given that #1558's E2E caught a defect the entire renderer suite was blind to, an E2E for the `#` menu is worth adding before merge rather than after.

## Out of scope

The journal keeps its own hand-rolled Tiptap `[[` suggestion; doing this there means writing it twice against two APIs, and it waits for #1547 (scope note added there). Junk-note cleanup #1555, snippet strippers #1556, CLI/MCP resolution #1557.

### DIFF
--- a/apps/desktop/src/renderer/src/components/note/content-area/ContentArea.test.tsx
+++ b/apps/desktop/src/renderer/src/components/note/content-area/ContentArea.test.tsx
@@ -26,6 +26,7 @@ const contentAreaMocks = vi.hoisted(() => ({
   fetchLinkPreview: vi.fn(),
   toastError: vi.fn(),
   defaultFileItemClick: vi.fn(),
+  openSuggestionMenu: vi.fn(),
   createLinkMentionContent: vi.fn(
     (url: string, domain: string, title?: string, favicon?: string) => ({
       type: 'linkMention',
@@ -344,6 +345,7 @@ function resetEditor(): void {
       if ('props' in update) block.props = { ...block.props, ...(update.props as object) }
     }),
     insertBlocks: vi.fn(),
+    getExtension: vi.fn(() => ({ openSuggestionMenu: contentAreaMocks.openSuggestionMenu })),
     getTextCursorPosition: vi.fn(() => ({ block: urlBlock })),
     prosemirrorView: { focus: vi.fn(), dom: { blur: vi.fn() } },
     _tiptapEditor: {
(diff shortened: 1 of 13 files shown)


h4yfans:
Consolidated into #1558 — everything on this branch is now on `fix/wiki-heading-links` as `ab52050d9` (the `#` picker and slash entry) and `4f3ece139` (editable links). One PR from here on, per Kaan. Closing; nothing is lost.

Issue #1612, opened 2026-08-19
CRDT: make the sweep's snapshot GET conditional, in-session — the warm sweep stops re-downloading every body


h4yfans:
Part of #1496. **PR 2 of 5** in the design merged as #1509. **Blocked by #1611** (the server must return `snapshotMeta` first).

> Spec: `git show 0b81f4293:docs/superpowers/specs/2026-08-16-crdt-sweep-catchup.md` — §2.1, §2.5, §2.6. It is not on disk; `cfb4fd22b` deleted `docs/superpowers`.

## Why

There are **two regimes** and today they cost the same ten minutes:

- **Warm** — this device has merged this vault's bodies before. Restart, reconnect after a Wi-Fi blip, the 15-minute interval sweep, the post-`fullSync` sweep. The bodies are already in the local Y.Doc store. Every one of the 1,000 snapshot GETs is redundant. **This is the overwhelmingly common case and it is 100% waste.**
- **Cold** — a new device, a rebuilt store, a restored vault. The transfer is real work.

Any design that does not separate them optimises the wrong thing. This PR separates them for the lifetime of one process — no persistence, so no exposure to FM2 (that is #1613's problem).

## What to build

Inside `applyCrdtBatchChunk` (`crdt-sync-coordinator.ts`), skip `applySnapshotBaseline` for a note when **both** hold:

1. `snapshotMeta[noteId].revision === merged[noteId].snapshotRevision`, **and**
2. `merged[noteId].appliedSequence >= snapshotMeta[noteId].sequenceNum`

`merged` is a **session-scoped in-memory map** in this PR (in-session `lastAppliedSequence` plus a revision map). Nothing on disk.

Fallbacks, all of which must default to *fetch*:

- `snapshotMeta` has **no entry** for the note → the server has no snapshot at all → skip trivially with `since = merged[noteId].appliedSequence ?? 0`.
- `snapshotMeta` is **`undefined` on the whole response** → the server is old → fetch, exactly as today.
- `merged[noteId] === undefined` → fetch.

## The two-phase sweep this unlocks

A note that skips its baseline **is never opened**. That matters: the 32-doc LRU cap on chunk size (`crdt-sync-coordinator.ts:261`) exists *only* because a chunk holds every one of its notes open across an await. A note with nothing to apply needs no doc, so the probe is not bound by the LRU at all.

[...]

Issue #1499, opened 2026-08-16
EPIC: CRDT/sync follow-ups from the 2026-08 offline data-loss investigation


h4yfans:
Residual work from the 2026-08-15/16 investigation into "an edit made while signed out disappears, including from the markdown file".

That bug is **fixed and manually verified** across all four combinations of the two devices being online/offline before and after. This epic holds what was deliberately left standing on the way there.

## Read this before triaging

These are **not unrelated cleanup.** Most are residual edges of the same sync system — real risks found while chasing the main bug and consciously not fixed, either because the fix needed its own evidence or because failing closed would have traded a rare loss for a certain one. Two of them are data-loss vectors. Please do not deprioritise the whole list on the assumption that it is hygiene.

## Data loss or corruption

- [x] #1489 — an update from an unresolvable signer can be pruned away by the next snapshot. **The one to do first.** The skipped update is deleted server-side while absent from the snapshot that replaced it.
- [x] #1490 — device linking rewrites the vault uuid and orphans that device's local CRDT store after the next restart.
- [x] #1491 — the legacy global store is inherited whole by whichever vault opens first; journal ids are deterministic, so the wrong history can attach. Pre-existing bleed, now frozen rather than recurring.
- [x] #1498 — open question: can bodies clobbered before #1449 be recovered from server history, and is anything owed to existing users?

## Correctness / robustness

- [x] #1492 — `teardownSession('integrity')` leaves the CRDT provider dead until restart. Likely a one-word fix plus a test.
- [x] #1494 — `drainPendingCrdtNotes` drops a concurrent drain instead of deferring it.
- [x] #1495 — the canvas note-card edit lock's premise no longer holds now that the local Y.Doc is live without a session. Either it is dead weight or its comment is wrong.

## Waste / performance

- [ ] #1493 — a never-signed-in device records pending note ids the first sync would push anyway.
- [ ] #1496 — a large vault takes minutes to catch up after sign-in. Deliberate pacing trade; shortening it is design work.

## Docs

- [x] #1497 — the sync-protocol endpoint table omits the CRDT snapshot and pull routes.

## Related, filed separately

- #1445 — a note over the 5MB encrypt cap never syncs and fails silently. Found in the same logs, different failure.

## What shipped (context for anyone picking this up)

Eight changes, all on `main`, in order. Each one was a real and distinct defect in the same scenario:

[...]

Issue #1497, opened 2026-08-16
docs: the sync-protocol endpoint table omits the CRDT snapshot and pull routes


h4yfans:
Split out of the 2026-08-15/16 sync investigation.

The Endpoints table in `apps/docs/src/architecture/sync-protocol.md` lists `POST /sync/crdt/updates` but not `POST /sync/crdt/snapshot`, `GET /sync/crdt/snapshot/:noteId`, `GET /sync/crdt/updates`, or `POST /sync/crdt/updates/batch`.

Those four carry every note body in the product and are now referenced by name in the surrounding prose, so the table is the odd one out. Small doc gap, noted while adding the snapshot broadcast section.

h4yfans:
Tracked under EPIC #1499 (CRDT/sync follow-ups from the 2026-08 offline data-loss investigation).

Issue #1496, opened 2026-08-16
CRDT: a large vault takes minutes to catch up after sign-in


h4yfans:
Split out of the 2026-08-15/16 sync investigation.

The vault-wide CRDT sweep is paced at 25 notes / 15s (`CRDT_SWEEP_CHUNK_NOTES`) so it fits inside the server's per-device budgets (300 GET + 30 batch POST per 60s). Notes with a live editor skip the queue, so what the user is looking at is immediate.

Consequence: a 1000-note vault takes roughly ten minutes to finish catching up after a sign-in or reconnect. Cost per minute is constant in vault size; only duration grows. That is the deliberate trade — a sweep is a catch-up, not a race — but it is a real wait for large vaults.

Two things would shorten it without breaking the budget:
- collapse the per-note snapshot GET (the batch endpoint batches incrementals but **not** snapshot baselines, so a sweep is still ~1 request per note)
- prioritise recently-opened notes after the live-editor ones

Both are design work, not tuning.

---

## Plan (added 2026-08-19)

The design landed as **#1509** and answers this issue: the sweep is slow because it spends **one snapshot GET per note per pass on a baseline it almost always already has**, not because the pacing is aggressive. The pacing runs at **one sixth** of the real GET budget.

Two regimes, and today they cost the same ten minutes:

- **Warm** — restart, reconnect, the 15-minute interval sweep. Bodies are already in the local Y.Doc store. All ~1,000 GETs are redundant. **The common case, and 100% waste** — a 1,000-note vault burns ~96,000 R2 Class B reads and ~288,000 D1 statements *per device per day* re-downloading bodies it has. That is a billing and battery issue, not just a wait.
- **Cold** — new device, rebuilt store, restored vault. The transfer is real work.

Target after the chain below: warm 1,000-note sweep goes from ~10 min / 1,000 GETs to **~40–60 s / zero snapshot GETs**; cold goes to ~3.5 min. Traffic goes *down* while the wall clock goes down ~10x.

### The five PRs

- [ ] **#1611 — PR 1 (server).** Snapshot `revision` token + `snapshotMeta` on the batch pull. Inert on its own; enables everything else. Respects the sync-server-before-desktop deploy order.
- [ ] **#1612 — PR 2 (client, in-memory).** Conditional snapshot skip inside `applyCrdtBatchChunk`. Fixes the reconnect and 15-minute sweeps within one process. *Blocked by #1611.*
[...]

PR #1617, opened 2026-08-19
feat(sync-server): add a snapshot revision token and return snapshotMeta on the batch CRDT pull

### LINKED ISSUE #1611: sync-server: add a snapshot `revision` token and return `snapshotMeta` on the batch CRDT pull
Part of #1496. **PR 1 of 5** in the design merged as #1509.

> The spec is `docs/superpowers/specs/2026-08-16-crdt-sweep-catchup.md`. It is **no longer on disk** — `cfb4fd22b` deleted `docs/superpowers` wholesale alongside an unrelated marketing-email edit. Recover it with:
> `git show 0b81f4293:docs/superpowers/specs/2026-08-16-crdt-sweep-catchup.md`
> Sections referenced below: §2.1, §2.4, §2.5, §2.6, §4.1.

## Why

The vault-wide sweep spends **one `GET /sync/crdt/snapshot/:noteId` per note per pass** on a baseline it almost always already has. `POST /sync/crdt/updates/batch` batches the *incrementals*; it does not batch the *baselines*. So a chunk of 25 costs 26 requests, and a 1,000-note vault burns ~1,000 GETs and ~1,000 R2 Class B reads every sweep — every 15 minutes, per device — to re-download bodies it already holds.

The client cannot currently know the GET is redundant, because it has no way to learn whether the **server's** snapshot changed without downloading it. This PR gives it that token.

**This PR is inert on its own.** No client behaviour changes. It is the enabling change, it lives entirely inside `apps/sync-server`, and it respects the fixed sync-server-before-desktop deploy order. It can sit deployed and unused indefinitely.

## What to build

1. **Migration `0005_crdt_snapshot_revision.sql`** — `ALTER TABLE crdt_snapshots ADD COLUMN revision TEXT NOT NULL DEFAULT ''`. Additive, with a default. **No backfill** (see below).
2. **`storeSnapshot`** assigns a fresh `crypto.randomUUID()` to `revision` on insert **and** in the `ON CONFLICT DO UPDATE SET` branch. Both paths, unconditionally.
3. **`getSnapshot`** returns `revision`, coalescing `''` at read time to `legacy:${id}:${created_at}:${size_bytes}`.
4. **New `getBatchSnapshotMeta`** reading `(note_id, sequence_num, revision, created_at, size_bytes, signer_device_id)` for the batch's note ids — folded into the existing `db.batch()` in `getBatchUpdates`, so it is one extra statement and **no extra round trip**.
5. **`handleCrdtBatchPull`** returns a new top-level `snapshotMeta` key; **`handleCrdtSnapshotPull`** returns `revision`.

Response shape (additive, one new top-level key):

\`\`\`jsonc
{
  "notes": { "<noteId>": { "updates": [...], "hasMore": false } },
  // NEW. Present on every response from a server that supports this.
  // A noteId absent from this map has no server snapshot at all.
  "snapshotMeta": {
    "<noteId>": { "sequenceNum": 42, "revision": "b1c9…", "signerDeviceId": "dev_…" }
  }
}
\`\`\`

## Why a random `revision` and not something we already have

**`sequence_num` cannot be the token.** `storeSnapshot` pins it: `sequenceNum = existingSnapshot?.sequence_num ?? currentSeq`. The second and every later snapshot for a note **keeps the same `sequence_num`** while replacing the blob. Using it would silently pin every device to the first snapshot a note ever had. This is the single easiest way to get this design wrong.

**`created_at` cannot be either** — 1-second resolution, and `(created_at, size_bytes, signer_device_id)` can collide for two pushes of the same note, in the same second, from the same device, at the same ciphertext length. The consequence of a collision is a permanently stale body.

**Random, not a counter**, so a row deleted and recreated (vault deletion, account recreation) cannot collide with a token a client still holds.

**No backfill migration.** Existing rows keep `''` and the server coalesces at read time. The legacy value is deterministic per row and changes on any rewrite — and any rewrite also assigns a real random revision, so a row is legacy at most until its next snapshot push. This avoids a single `UPDATE` over what could be millions of rows inside a D1 migration.

**Rejected zero-migration alternative** (recorded so it is not re-proposed): add `id = excluded.id` to the `DO UPDATE SET` clause and reuse `id` as the revision, since it is already a fresh `crypto.randomUUID()` per store and nothing references it. Rejected because it silently changes `id` from a row identity into a version token, which the next reader of that table will not expect.

## Failure mode this PR owns

**FM1 — the token fails to change when the blob does.** A client then skips a snapshot it needed and keeps a stale body forever. This is the whole design's central risk and it is created here.

**Mitigation is a test requirement, not care:** the revision-bump test must be **mutation-verified** — delete the revision bump from the `ON CONFLICT` branch and the test has to go red. Green unit tests over broken behaviour is the documented failure mode of this subsystem (#1499).

## Compat

- **New server + old client** — this is the window the deploy order creates. Request shape unchanged. `snapshotMeta` and `revision` are additive response fields. The desktop client reads batch responses through \`postToServer<CrdtBatchPullResponse>\`, a TypeScript cast with **no runtime schema validation** (the server's zod covers requests only), so extra keys are ignored. Verified in the spec, not assumed.
- **D1** — additive column with a default; old server code never reads or writes it.
- **Rate-limit classes** — unchanged. No new bucket, no changed limit.

## Verification

- \`apps/sync-server/src/services/crdt.test.ts\` — revision assigned on insert **and** on conflict; mutation-verified per FM1.
- \`apps/sync-server/src/schema/d1.test.ts\` — the \`PRAGMA table_info\` column-set assertion (~\`:188\`) asserts the old column set and **must be updated in this PR**.
- \`pnpm test:sync-server\` · \`pnpm typecheck\` · \`pnpm lint\`
- Docs: \`apps/docs/src/architecture/sync-protocol.md\` — add \`snapshotMeta\` and \`revision\` to the endpoint table (also closes part of #1497).

## Honest weakness

On its own **a user feels nothing.** If the goal is one PR a user can feel, do the ordering issue instead — it is independent and can land at any point.

h4yfans:
**PR 1 of 5** in #1496. #1612 stacks on this branch.

## What changed

The vault sweep spends one unconditional `GET /sync/crdt/snapshot/:noteId` per note per pass on a baseline it almost always already holds — ~1,000 GETs and ~1,000 R2 Class B reads every 15 minutes on a 1,000-note vault. The client cannot know the GET is redundant, because it has no way to learn whether the **server's** snapshot changed without downloading it. This PR adds that token.

- **`migrations/0005_crdt_snapshot_revision.sql`** — `ALTER TABLE crdt_snapshots ADD COLUMN revision TEXT NOT NULL DEFAULT ''`. Additive, no backfill.
- **`storeSnapshot`** assigns a fresh `crypto.randomUUID()` to `revision` on insert **and** in the `ON CONFLICT DO UPDATE SET` branch — both paths, unconditionally.
- **`getSnapshot`** returns `revision`, coalescing `''` at read time to `legacy:${id}:${created_at}:${size_bytes}`.
- **`getBatchSnapshotMeta`** folds one extra statement into the `db.batch()` `getBatchUpdates` already sends — one extra statement, **no extra round trip**. `getBatchUpdates` now returns `{ notes, snapshotMeta }`.
- **`POST /sync/crdt/updates/batch`** gains a top-level `snapshotMeta` map; **`GET /sync/crdt/snapshot/:noteId`** gains `revision`.

```jsonc
{
  "notes": { "<noteId>": { "updates": [...], "hasMore": false } },
  // NEW. Present on every response from a server that supports this.
  // A noteId absent from this map has no server snapshot at all.
  "snapshotMeta": {
    "<noteId>": { "sequenceNum": 42, "revision": "b1c9…", "signerDeviceId": "dev_…" }
  }
}
```

**This PR is inert.** No client reads either field yet. It lives entirely inside `apps/sync-server` and respects the fixed sync-server-before-desktop deploy order; it can sit deployed and unused indefinitely.

### One deviation from the issue, called out

The issue lists the batch metadata read as `(note_id, sequence_num, revision, created_at, size_bytes, signer_device_id)`. It also selects `id`, because the legacy coalesce is `legacy:${id}:…` and **both read paths have to produce the same string for the same row**. Without `id` in the batch select, a client that merged `legacy:<id>:…` from the GET would compare it against `legacy:<note_id>:…` from the batch, never match, and re-download every pre-migration snapshot forever. A single shared `coalesceRevision` helper now serves both paths, and a test asserts they agree.

## Why not something we already have

**`sequence_num` cannot be the token.** `storeSnapshot` pins it (`existingSnapshot?.sequence_num ?? currentSeq`) so later incrementals stay pullable — the second and every later snapshot keeps the same number while replacing the blob. Using it would silently pin every device to the first snapshot a note ever had. The FM1 test asserts the revision moves *while the sequence number does not*, in the same test, so a future "simplification" to `sequence_num` fails loudly.

**`created_at` cannot be either** — 1-second resolution, and `(created_at, size_bytes, signer_device_id)` can collide for two pushes of the same note, in the same second, from the same device, at the same ciphertext length. The consequence of a collision is a permanently stale body.

**Random, not a counter**, so a row deleted and recreated (vault deletion, account recreation) cannot collide with a token a client still holds.

**Rejected zero-migration alternative**, recorded so it is not re-proposed: reuse `id` as the revision via `id = excluded.id`. Rejected because it silently changes `id` from a row identity into a version token, which the next reader of that table will not expect.

## Compat plan

- **PRODUCTION, real users.** Additive column **with a default** and **no backfill `UPDATE`** over the table — this avoids a single `UPDATE` over what could be millions of rows inside a D1 migration. Existing rows keep `''` and the server coalesces at read time; the legacy token is deterministic per row and changes on any rewrite, and any rewrite also assigns a real random revision, so a row is legacy at most until its next snapshot push.
- **New server + old client** (the window the deploy order creates). Request shape unchanged. `snapshotMeta` and `revision` are additive response fields; the desktop client reads batch responses through `postToServer<CrdtBatchPullResponse>`, a TypeScript cast with no runtime schema validation (the server's zod covers requests only), so extra keys are ignored.
- **Old server code** never reads or writes the column, so a rollback after the migration is a no-op.
- **Rate-limit classes** unchanged. No new bucket, no changed limit.
- `GET /sync/crdt/snapshot/:noteId` returns `revision: null` on its no-snapshot branch, matching the existing `signerDeviceId: null`, so the response shape is uniform.

## FM1 — mutation evidence

**FM1: the token fails to change when the blob does.** A client then skips a snapshot it needed and keeps a stale body forever. The `ON CONFLICT` branch is where it goes wrong, because the row already exists and `sequence_num` is pinned across it.

The fake D1 in `crdt.test.ts` was changed to model `DO UPDATE SET` **column by column** rather than overwriting the whole row — without that, the mutation below would not have been observable at all, and the test would have been green over broken behaviour (the documented failure mode of this subsystem, #1499).

Mutation applied (`git diff --stat`: `apps/sync-server/src/services/crdt.ts | 2 +-`):

```diff
-         DO UPDATE SET blob_key = excluded.blob_key, sequence_num = excluded.sequence_num, size_bytes = excluded.size_bytes, signer_device_id = excluded.signer_device_id, created_at = excluded.created_at, revision = excluded.revision`
+         DO UPDATE SET blob_key = excluded.blob_key, sequence_num = excluded.sequence_num, size_bytes = excluded.size_bytes, signer_device_id = excluded.signer_device_id, created_at = excluded.created_at`
```

Result — exactly the FM1 test goes red, nothing else:

```
 × assigns a NEW revision when a snapshot is replaced, even though the sequence stays pinned 3ms
AssertionError: expected '6d60d664-5c1e-430a-98c1-b0dcf484ccf1' not to be '6d60d664-5c1e-430a-98c1-b0dcf484ccf1'
 Test Files  1 failed (1)
      Tests  1 failed | 18 passed (19)
```

The mutation was then reverted and the full suite re-run green.

## Verification

| gate | result |
| --- | --- |
| `pnpm test:sync-server` | **PASS** — 62 files, 963 tests, 0 failed |
| `pnpm lint` | **PASS** — 0 errors, 97 warnings, none in `apps/sync-server` |
| `git diff --check` | **PASS** — clean |
| `pnpm docs:impact --base 6cc0dd83f --strict` | **PASS** — `covered` |
| `pnpm docs:build` | **PASS** |
| `pnpm typecheck` | **FAIL, pre-existing** — see below |

`pnpm typecheck` fails with one error, `apps/sync-server/src/services/posthog-transform.test.ts(315,9): error TS2741: Property 'failure' is missing…`. It is not from this branch: that file is byte-identical to the base commit (`git diff --quiet 6cc0dd83f -- <file>` is clean), and the fix for exactly this error is 81354bfa2 ("fix(sync): stop re-asking the server for attachments it does not have", #1598), which is on `main` but landed after this branch's base. It resolves on rebase or merge. No other typecheck errors anywhere in the repo.

Docs updated in `apps/docs/src/architecture/sync-protocol.md` (endpoint table + a new "Snapshot revisions" section), which also closes part of #1497.

Closes #1611.

### DIFF
--- a/apps/docs/src/architecture/sync-protocol.md
+++ b/apps/docs/src/architecture/sync-protocol.md
@@ -548,9 +548,9 @@ list.
 | `POST /sync/pull`                 | down      | Fetch updates since cursor                                                       |
 | `POST /sync/crdt/updates`         | up        | Incremental Yjs binary updates                                                   |
 | `GET /sync/crdt/updates`          | down      | One note's incremental updates (`note_id`, `since`, `limit` query params)        |
-| `POST /sync/crdt/updates/batch`   | down      | Incremental updates for up to 100 notes in one request; never snapshot baselines |
+| `POST /sync/crdt/updates/batch`   | down      | Incremental updates for up to 100 notes in one request, plus `snapshotMeta`      |
 | `POST /sync/crdt/snapshot`        | up        | Full Yjs document baseline; prunes the note's stored updates at or below it      |
-| `GET /sync/crdt/snapshot/:noteId` | down      | The note's snapshot baseline, applied before its incrementals                    |
+| `GET /sync/crdt/snapshot/:noteId` | down      | The note's snapshot baseline and its `revision`, applied before its incrementals |
 | `GET /sync/vaults`                | down      | List the account's registered vaults                                             |
 | `POST /sync/vaults`               | up        | Register or update a vault's encrypted name                                      |
 | `POST /auth/*`                    | mixed     | OTP, sign-in, refresh, sign-out                                                  |
@@ -564,6 +564,31 @@ moves metadata only. A device reads a body by applying the baseline from
 `POST /sync/crdt/updates/batch` is the whole-vault form of that second step — up to 100 notes per
 request, each with its own `since` — but it batches incrementals only, so the baselines stay one GET
 per note and dominate a first-sync sweep's request count.
+
+### Snapshot revisions
+
+A snapshot baseline carries a `revision`: an opaque token the server replaces on every snapshot
+write, so a client can tell whether the server's baseline moved without downloading it.
+
+- `GET /sync/crdt/snapshot/:noteId` returns `revision` alongside `snapshot`, `sequenceNum` and
+  `signerDeviceId`, so a device learns the token for the baseline it just merged. When the note has
+  no server snapshot the response is `{ snapshot: null, sequenceNum: 0, signerDeviceId: null, revision: null }`.
+- `POST /sync/crdt/updates/batch` returns a top-level `snapshotMeta` map beside `notes`:
+  `{ "<noteId>": { "sequenceNum": 42, "revision": "…", "signerDeviceId": "…" } }`. A note absent from
+  a present map has no server snapshot at all; an absent `snapshotMeta` key means the server predates
+  this field.
+
+The token is deliberately not `sequenceNum`. A replacement snapshot keeps the note's existing
+sequence number so later incrementals stay pullable, so the number does not move when the blob does.
+It is random per write rather than a counter, so a row deleted and recreated — vault deletion,
+account recreation — cannot reuse a token a client still holds.
+
+Rows written before the field existed are not backfilled; the server derives a deterministic token
+for them at read time from the row's identity, creation time and size, and the next snapshot push
+replaces it with a real one. Both read paths return the same token for the same row.
+
+Both fields are additive: an older client reads these responses through an unvalidated cast and
+ignores the extra keys.
 
 Pushing a snapshot is an assertion, not just a write: once the snapshot is stored,
 `pruneUpdatesBeforeSnapshot` deletes every `crdt_updates` row for that note at or below the
--- a/apps/sync-server/migrations/0005_crdt_snapshot_revision.sql
+++ b/apps/sync-server/migrations/0005_crdt_snapshot_revision.sql
@@ -1,0 +1,31 @@
+-- A change token for the snapshot blob, so a client can learn whether the
+-- SERVER's snapshot moved without downloading it. The vault sweep currently
+-- spends one unconditional `GET /sync/crdt/snapshot/:noteId` per note per pass
+-- on a baseline it almost always already holds; this column is what makes that
+-- GET conditional.
+--
+-- Nothing we already store can play this role:
+--
+--   * `sequence_num` is PINNED on rewrite -- `storeSnapshot` keeps
+--     `existingSnapshot?.sequence_num` so later incrementals stay pullable, so
+--     the second and every later snapshot for a note keeps the same number
+--     while replacing the blob. Using it would pin every device to the first
+--     snapshot a note ever had.
+--   * `created_at` has 1-second resolution, and (created_at, size_bytes,
+--     signer_device_id) can collide for two pushes of the same note, in the
+--     same second, from the same device, at the same ciphertext length. The
+--     consequence of a collision is a permanently stale note body.
+--
+-- Random per write rather than a counter, so a row deleted and recreated (vault
+-- deletion, account recreation) cannot collide with a token a client still holds.
+--
+-- Additive with a default and deliberately NOT backfilled. '' is the signal that
+-- a row predates this column; `getSnapshot` and the batch metadata read coalesce
+-- it to `legacy:<id>:<created_at>:<size_bytes>`, which is deterministic per row
+-- and changes on any rewrite -- and any rewrite also assigns a real random
+-- revision, so a row is legacy at most until its next snapshot push. That avoids
+-- a single UPDATE over what could be millions of rows inside a D1 migration.
+--
+-- A Worker deployed before this migration (or rolled back after it) never reads
+-- or writes the column, so old server code is unaffected.
+ALTER TABLE crdt_snapshots ADD COLUMN revision TEXT NOT NULL DEFAULT '';
--- a/apps/sync-server/schema/d1.test.ts
+++ b/apps/sync-server/schema/d1.test.ts
@@ -199,6 +199,7 @@ describe('D1 schema', () => {
     expect(cursorColumns.map((column) => column.name)).toContain('vault_id')
     expect(crdtUpdateColumns.map((column) => column.name)).toContain('vault_id')
     expect(crdtSnapshotColumns.map((column) => column.name)).toContain('vault_id')
+    expect(crdtSnapshotColumns.map((column) => column.name)).toContain('revision')
     expect(uploadSessionColumns.map((column) => column.name)).toContain('vault_id')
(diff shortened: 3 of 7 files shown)


h4yfans:
Rebased onto current `main` (`3eaba3c8f`) and force-pushed. The rebase was clean.

This supersedes the "typecheck fails, pre-existing" note in the PR body: the single `posthog-transform.test.ts` error came from `81354bfa2` (#1598) having landed *after* this branch's original base. With that upstream commit in, the gate is fully green.

Re-verified on the rebased tip (`fa930715b`):

| gate | result |
| --- | --- |
| `pnpm typecheck` | **PASS** — exit 0, **0 TS errors** repo-wide |
| `pnpm test:sync-server` | **PASS** — 62 files, **966 tests, 0 failed** |

Stack position: this is PR 1 of 5 for #1496. #1612 is branched off this branch and its PR targets `feat/crdt-snapshot-revision-meta`, so **merge this one first**; the child will retarget to `main` automatically once this lands.

PR #1618, opened 2026-08-19
perf(sync): order the CRDT catch-up sweep — open docs first, then modifiedAt DESC

### LINKED ISSUE #1614: CRDT: order the paced sweep — open docs first, then `modifiedAt DESC`
Part of #1496. **PR 5 of 5** in the design merged as #1509 — but **independent of all the others and can land at any point, including first.** No blockers.

> Spec: `git show 0b81f4293:docs/superpowers/specs/2026-08-16-crdt-sweep-catchup.md` — §3. Not on disk; `cfb4fd22b` deleted `docs/superpowers`.

## Why this one is cheap

This is an **ordering change at the source and nothing else.** Order is already preserved end to end: `pendingPulls` is a `Set`, `drainPendingPulls` is `Array.from` on it, `pacedCrdtPullQueue` is a `Set`, and the chunk is taken by iterating it. **Insertion order *is* priority.** Nothing else has to change to make ordering work.

Today tier 1 already exists — `flushPendingCrdtPulls` (`full-sync-runner.ts:401-431`) splits the drained set into notes with a **live editor**, which bypass the pace entirely, and everything else. Everything else then arrives in whatever order `getAllCrdtNoteIds` returns, which is an unordered `SELECT id FROM note_cache WHERE file_type = 'markdown'` (`note-crud.ts:277-284`) — effectively rowid order, i.e. **index-build order**.

## What to build

Three tiers instead of two:

1. **Live editors** — as today, un-paced. No change.
2. **Open-but-inactive docs** — `crdtProvider.getOpenNoteIds()` minus the active set, at the **front** of the paced queue. This is free and it is the best available proxy: the 32-doc LRU **is** a recently-opened list of exactly 32 notes, already in memory, already the notes a user is one click from. Today they get no priority at all.
3. **The rest of the vault, ordered `modifiedAt DESC`** — one `ORDER BY` on `getAllCrdtNoteIds`. `note_cache.modifiedAt` already exists (`note-crud.ts:290` uses it). A note that changed recently — by this user or by the device we are catching up with — is both the one most likely to be opened next and the one most likely to actually be stale.

Files: `note-crud.ts` and `full-sync-runner.ts`. That is the whole surface.

**Rejected:** `searchReasons.visitedAt` as the recency source — it is search-only, capped at 20 rows, and lives in the data DB (`ipc/search-handlers.ts:132-207`). Not worth reaching for over tier 2.

## The invariant that must survive

**Priority must never become filtering.** The sweep is the only channel by which a body-only remote edit reaches a device that missed the broadcast — note bodies never travel in the record change feed (`full-sync-runner.ts:112-128`). Skipping notes by heuristic converts a slow catch-up into a **silent** one.

The sweep stays exhaustive; only the order changes. Any implementation must preserve that, and it should be asserted in a test, not assumed.

## Failure modes

- **A vault whose mtimes are uniform** — restored from backup, freshly cloned, bulk-imported. Ordering degrades to arbitrary, i.e. **exactly today's behaviour**. No regression.
- **Failed notes go to the back.** `owePendingPull` re-adds to `pendingPulls`, so a re-queued note lands at the end of the *next* drain. That is correct and **should stay**: a note that just failed is the worst candidate for a front-of-queue retry.

## Arithmetic

**None.** Zero change to request count, request rate, or duration. Every rate-limit budget is untouched.

## Compat

None needed. Client-only. No protocol, no schema, no persisted state, no migration.

## Honest assessment

This changes **perceived** latency and nothing else. On its own it does not shorten the ten minutes, does not reduce cost, and does not remove one redundant request.

After #1612 / #1613 land, the warm sweep finishes in under a minute, so ordering stops mattering there and matters only for **cold** catch-up — where it matters a lot, because cold is where the ~3.5 minutes of genuine body transfer sits.

It is also the cheapest change in the whole design by a wide margin, and the only one a user can feel without a server deploy. If the goal is one small independent win first, this is it.

## Verification

- `pnpm --filter @memry/desktop test:main` · `pnpm typecheck` · `pnpm lint`
- Test that the sweep still visits **every** note (the invariant above).
- Test tier ordering: open-but-inactive docs precede the `modifiedAt DESC` tail.

h4yfans:
Closes #1614. PR **5 of 5** in the design merged as #1509 (epic #1496) — but **independent of the other four**. No blockers, no ordering constraint: it can merge at any point, including first.

## What changed

The paced CRDT sweep drains FIFO, so the order its work list arrives in *is* the order a user watches stale bodies get repaired in. Today that order comes from an unordered `SELECT id FROM note_cache WHERE file_type = 'markdown'` — rowid order, i.e. index-build order, which correlates with nothing.

Three tiers instead of two:

1. **Live editors** — unchanged. Still pulled in their own batch outside the pace.
2. **Open-but-inactive docs** — `crdtProvider.getOpenNoteIds()` minus the active set, spliced in at the **front** of the paced queue. The 32-doc LRU already is a recently-opened list, in memory and free to read; those are the notes the user is one click from, and today they get no priority at all. Front-inserted rather than appended, because a sweep landing mid-drain would otherwise put them behind everything the previous pass left waiting — which on a large vault is minutes.
3. **The rest of the vault, `modifiedAt DESC`** — one `ORDER BY` on `getAllCrdtNoteIds`, covered by the existing `idx_note_cache_modified`.

Nothing else had to change: `pendingPulls` is a `Set`, `drainPendingPulls` is `Array.from` on it, `pacedCrdtPullQueue` is a `Set`, and the chunk is taken by iteration. Insertion order already *is* priority end to end — verified in the code, and now pinned by a test.

Files: `note-crud.ts`, `full-sync-runner.ts`. That is the whole surface.

## Why the ordering is safe

**Priority must never become filtering.** The vault-wide sweep is the only channel by which a body-only remote edit reaches a device that missed the `crdt_updated` broadcast — note bodies never travel in the record change feed (`NoteSync` sends `content: null` on update; see the invariant comment at `full-sync-runner.ts:112-128`). A note skipped by a heuristic would keep a stale body with nothing else to discover the edit, turning a slow catch-up into a **silent** one.

So this is a reordering and nothing else, and that is asserted rather than assumed:

- `full-sync-runner.test.ts` → `#then prioritising never drops a note from the sweep`: a 60-note vault with both an active editor and an open-but-inactive doc drains completely — `new Set(pulled)` equals the full vault, and `pulled` has length 60, so exactly once, no duplicates either.
- `crdt-sweep-order.test.ts` → `still returns every markdown note, whatever the order`: the query itself stays exhaustive at the source.

**Failed notes still go to the back.** `owePendingPull` re-adds to `pendingPulls`, so a re-queued note lands at the end of the *next* drain. Deliberately left alone — a note that just failed is the worst candidate for a front-of-queue retry.

**Not** `searchReasons.visitedAt` as the recency source: search-only, capped at 20 rows, different DB.

## Failure modes

- **Uniform mtimes** — a vault restored from backup, freshly cloned or bulk-imported. Ordering degrades to arbitrary, i.e. **exactly today's behaviour**. No regression. Covered by `keeps notes whose mtimes are identical rather than collapsing them`.
- **Correctness unaffected** — the sweep stays exhaustive; only positions change.
- **`ORDER BY` must not widen the `WHERE`** — a PDF in the sweep is a pull for a doc that does not exist. Covered by `leaves non-markdown files out, as before`.

## Arithmetic and compat

**None, and none.** Zero change to request count, request rate or duration; every rate-limit budget is untouched. Client-only: no protocol, no schema, no persisted state, no migration. This changes *perceived* latency and nothing else.

## Tests added (9)

`apps/desktop/src/main/sync/engine/full-sync-runner.test.ts`
- open-but-inactive docs lead the paced queue
- a live editor still outranks an open-but-inactive doc
- the order the sweep supplies is the order the queue drains (pins the end-to-end order preservation the whole change rests on)
- open docs jump ahead of a drain already in progress
- prioritising never drops a note from the sweep

`apps/desktop/src/main/database/queries/notes/crdt-sweep-order.test.ts` (new)
- returns the most recently modified note first (inserted oldest-first, so rowid order is the exact opposite of the answer)
- still returns every markdown note, whatever the order
- keeps notes whose mtimes are identical rather than collapsing them
- leaves non-markdown files out, as before

New test file rather than an addition to `notes.test.ts` because that file is on the `tsconfig.test.*` exclude backlog, and the backlog only shrinks.

## Verification

| Gate | Result |
| --- | --- |
| `pnpm --filter @memry/desktop test:main` | **PASS** — 541 files, 7032 passed, 1 expected fail, 6 skipped |
| `pnpm lint` | **PASS** — 0 errors (97 pre-existing warnings in untouched files); changed source files clean |
| `pnpm exec prettier --check` on all 5 changed files | **PASS** |
| `git diff --check` | **PASS** |
| `pnpm docs:impact --base 6cc0dd83f --strict` | **PASS** — `covered` |
| `pnpm docs:build` | **PASS** |
| `pnpm typecheck` | **FAIL — pre-existing, not this branch** (see below) |

`pnpm typecheck` fails on two files this branch does not touch, both already red at the base commit `6cc0dd83f`:

- `apps/desktop/src/renderer/src/components/sidebar/sidebar-nav.test.tsx` — 2× `TS2741: Property 'onNavMiddleClick' is missing`
- `apps/sync-server/src/services/posthog-transform.test.ts` — 1× `TS2741: Property 'failure' is missing`

`git status` on this branch lists only the 5 files in this diff; neither of those appears. Both have since been fixed on `origin/main` (this branch is based on `6cc0dd83f`, which `origin/main` has moved past), so they resolve on merge.

The `test:main` number above was taken **after** `bash apps/desktop/scripts/ensure-native.sh node`, so `better-sqlite3` was on the Node ABI (the worktree install's postinstall only builds the Electron ABI, and a wrong ABI can return quietly wrong SQLite answers rather than failing loudly). The result was byte-identical before and after the rebuild — 541 files / 7032 passed / 1 expected fail / 6 skipped both times — so nothing was masked, and the four new SQLite-backed `crdt-sweep-order.test.ts` cases pass on the correct ABI.

Also re-verified: `crdt-sync-coordinator.test.ts` failed once on the very first full run with `Error: Electron failed to install correctly` — a race against the worktree's `electron-rebuild` postinstall, not a test failure. It passes standalone (36/36) and in both clean full runs above.

## Docs

`apps/docs/src/architecture/crdt.md` gains a **Sweep priority** subsection under *Pacing the sweep*: the three tiers, why tier 2 is front-inserted, the exhaustiveness invariant, the failed-notes-to-the-back rule, and the uniform-mtime degradation.

### DIFF
--- a/apps/desktop/src/main/database/queries/notes/crdt-sweep-order.test.ts
+++ b/apps/desktop/src/main/database/queries/notes/crdt-sweep-order.test.ts
@@ -1,0 +1,90 @@
+import { afterEach, beforeEach, describe, expect, it } from 'vitest'
+import type { TestDatabaseResult, TestDb } from '@tests/utils/test-db'
+import { createTestIndexDb } from '@tests/utils/test-db'
+import type { NewNoteCache } from '@memry/db-schema/schema/notes-cache'
+import { getAllCrdtNoteIds, insertNoteCache } from '.'
+
+// getAllCrdtNoteIds is the source of the CRDT catch-up sweep's work list, and
+// the order it returns is the order the sweep pulls in — the paced queue drains
+// FIFO, so this SELECT decides which stale note a user sees fixed first. It used
+// to be unordered, i.e. rowid/index-build order, which correlates with nothing.
+// #1614 makes it modifiedAt DESC. The order is a priority and must never become
+// a filter: the sweep is the only channel by which a body-only remote edit
+// reaches a device that missed the `crdt_updated` broadcast.
+
+describe('getAllCrdtNoteIds — sweep priority order', () => {
+  let dbResult: TestDatabaseResult
+  let db: TestDb
+
+  const createNote = (id: string, overrides: Partial<NewNoteCache> = {}): void => {
+    insertNoteCache(db, {
+      id,
+      path: overrides.path ?? `notes/${id}.md`,
+      title: overrides.title ?? `Note ${id}`,
+      contentHash: `hash-${id}`,
+      createdAt: '2026-01-10T00:00:00.000Z',
+      modifiedAt: overrides.modifiedAt ?? '2026-01-12T00:00:00.000Z',
+      ...(overrides.fileType !== undefined && { fileType: overrides.fileType })
+    })
+  }
+
+  beforeEach(() => {
+    dbResult = createTestIndexDb()
+    db = dbResult.db
+  })
+
+  afterEach(() => {
+    dbResult.close()
+  })
+
+  it('returns the most recently modified note first', () => {
+    // Inserted oldest-first so rowid order is the exact opposite of the answer:
+    // an unordered SELECT passes only by accident, and not this way round.
+    createNote('oldest', { modifiedAt: '2026-01-01T00:00:00.000Z' })
+    createNote('middle', { modifiedAt: '2026-02-01T00:00:00.000Z' })
+    createNote('newest', { modifiedAt: '2026-03-01T00:00:00.000Z' })
+
+    expect(getAllCrdtNoteIds(db)).toEqual(['newest', 'middle', 'oldest'])
+  })
+
+  it('still returns every markdown note, whatever the order', () => {
+    // The invariant. Ordering may shuffle the work list; it may never shorten
+    // it. A note dropped here keeps a stale body with no other path to discover
+    // the remote edit, which turns a slow catch-up into a silent one.
+    const ids = Array.from({ length: 40 }, (_, index) => `note-${index}`)
+    for (const [index, id] of ids.entries()) {
+      createNote(id, {
+        modifiedAt: `2026-01-${String((index % 28) + 1).padStart(2, '0')}T00:00:00.000Z`
+      })
+    }
+
+    const swept = getAllCrdtNoteIds(db)
+
+    expect(swept).toHaveLength(ids.length)
+    expect(new Set(swept)).toEqual(new Set(ids))
+  })
+
+  it('keeps notes whose mtimes are identical rather than collapsing them', () => {
+    // A vault restored from backup, freshly cloned or bulk-imported has uniform
+    // mtimes. Ordering degrades to arbitrary — exactly the behaviour before the
+    // ORDER BY — but every note must still be swept.
+    createNote('a', { modifiedAt: '2026-01-12T00:00:00.000Z' })
+    createNote('b', { modifiedAt: '2026-01-12T00:00:00.000Z' })
+    createNote('c', { modifiedAt: '2026-01-12T00:00:00.000Z' })
+
+    expect(getAllCrdtNoteIds(db).sort()).toEqual(['a', 'b', 'c'])
+  })
+
+  it('leaves non-markdown files out, as before', () => {
+    // Only markdown notes are CRDT-backed. The ORDER BY must not widen the
+    // WHERE: a PDF in the sweep is a pull for a doc that does not exist.
+    createNote('note', { modifiedAt: '2026-01-01T00:00:00.000Z' })
+    createNote('scan', {
+      fileType: 'pdf',
+      path: 'notes/scan.pdf',
+      modifiedAt: '2026-06-01T00:00:00.000Z'
+    })
+
+    expect(getAllCrdtNoteIds(db)).toEqual(['note'])
+  })
+})
--- a/apps/desktop/src/main/database/queries/notes/note-crud.ts
+++ b/apps/desktop/src/main/database/queries/notes/note-crud.ts
@@ -274,11 +274,30 @@ export function getAllNoteIds(db: IndexDb): string[] {
     .map((r) => r.id)
 }
 
+/**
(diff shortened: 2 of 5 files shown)


h4yfans:
Rebased onto current `main` (`3eaba3c8f`) and force-pushed. Clean rebase.

This supersedes the "typecheck fails, pre-existing" note in the PR body. Both errors (`sidebar-nav.test.tsx` ×2, `posthog-transform.test.ts` ×1) were already fixed upstream after this branch's original base, so they resolve with the rebase rather than needing anything here.

Re-verified on the rebased tip (`1cb9f6b1c`):

| gate | result |
| --- | --- |
| `pnpm typecheck` | **PASS** — exit 0, **0 TS errors** repo-wide |
| `pnpm --filter @memry/desktop test:main` | **PASS** — 543 files, **7079 passed**, 1 expected fail, 6 skipped |

Stack note: this is PR 5 of 5 for #1496 but it is **independent of the other four** — no protocol, schema, persisted state or request-rate change. It can merge at any point, including first.

PR #1635, opened 2026-08-20
feat(home): add a "Recently opened" widget beside "Recently edited"

h4yfans:
## Summary

A note you open but never edit surfaces nowhere. The existing Home widget sorts by `note_cache.modified_at`, so read-only reference notes are invisible on the dashboard. The report behind this framed it as a naming problem — a user read "Recently Edited" as "Recently opened" — but the label already says "Recently edited" and every row subtitle already says "edited {time}". The misread was an expectation gap, not a wording one, so **nothing is renamed here**; the missing widget is the fix. The two now sit side by side, and the row subtitle ("opened 12m ago" vs "edited 3m ago") is what tells them apart.

**The trail is device-local, on purpose.** It is not a sync item type: a write per note activation is a far higher push rate than per edit, and the CRDT sweep is already pacing against 429s. Board layout still syncs unchanged through the existing `home_page` item, so the widget's position travels between devices with **no new sync code**.

What landed:

- **`recently_opened`** (data DB) stores `{ item_id, item_type, opened_at }` only — one upserted row per note, pruned to 50. Title and emoji resolve from the note cache at read time, so a renamed note reads correctly and a deleted one drops out, rather than lingering under the stale denormalized title `search_reasons` keeps.
- **Migration 0053** seeds once from `search_reasons` so the widget is not born empty on an existing install. A copy, not a link — the tables diverge from there.
- **The signal is renderer-side**: the active note tab must hold focus for 2s before it counts, and repeat visits inside a minute do not write again. `note_opened` telemetry cannot stand in — it fires on every `notes:get`, including react-query refetches after a save or a peer sync.
- **`item_type`** is carried but always `'note'` for now; canvas and journal can join later without a schema migration.
- **Not added to the default board seed** — a fresh device has no history, so seeding it would greet every new install with an empty card. Available from the gallery.

### Reviewer notes

- **Backward compat.** `WidgetFrame`'s `onRemove` is now optional and omitted for a widget type this build does not know. Boards are whole-row LWW, so before this an older app pulling a newer widget could delete it out from under the device that added it — a silent undo, not a crash. The guard applies to every widget added from here on.
- **Two DBs, no join.** The trail lives in the data DB and `note_cache` in the index DB, so resolution happens in code rather than SQL. The trail deliberately does *not* live in the index DB: that one is a rebuildable cache, and a reindex would wipe the history.
- **Known pre-existing red, untouched:** `fts-corruption.test.ts` → "cannot empty a corrupt table" fails on `main` too; nothing in this diff reaches search/FTS.

## Release note

Home has a new **Recently Opened** widget for the notes you read but never edited — a companion to Recently Edited. Its list stays on the device that made it; the widget itself moves between machines like any other.

## Test plan

- `pnpm --filter @memry/desktop test:main` — new `recently-opened` query suite 6/6 (ordering, one-row-per-note upsert, live title resolution, deleted-note drop, retention prune, caller limit)
- `pnpm exec vitest --project renderer` (full) — 7714 passed; new hook suite covers dwell, tab-flip-past, throttle, post-window re-record, non-note tabs
- home renderer suite 83/83 · migration suites 15/15 · `ipc/index.test.ts` 4/4
- `pnpm typecheck` (web/node/test) · `pnpm lint` · `pnpm ipc:check` · `pnpm check:architecture` · `pnpm check:contracts` · `i18n:check` — all green
- `pnpm docs:impact --base <merge-base> --strict` · `pnpm docs:build` — green
- Not done: no manual run of the packaged app. The dwell/throttle signal is exercised by unit tests with fake timers, but a real open-a-note pass is worth doing before this leaves draft.


h4yfans:
### Migration renumbered to 0054 — it was being silently skipped

Caught while dogfooding: `recents:list` failed at runtime with `SqliteError: no such table: recently_opened`.

The branch minted `0053_recently_opened` with `when` hand-picked as 0052+1ms (`1787168001775`). main then shipped **`0053_custom_icons` with the identical value** — same synthetic step off the same predecessor. Drizzle applies a migration iff `lastApplied < folderMillis` and never compares hashes, so on any vault that had already taken main's 0053 the watermark landed *exactly* on ours, the `<` test failed, and `CREATE TABLE recently_opened` was skipped with no error at all.

Fixed by merging main and renumbering to **`0054_recently_opened`** with `when` from a real `Date.now()` (`1787226759433`), above main's tip. Verified by replaying the migration folder against a copy of a real dev vault that was already stuck at the colliding watermark — the table is created and the one-time `search_reasons` hand-off seeds it.

Worth flagging for reviewers: `migrate-journal.test.ts` could not have caught this. It asserts the newest journal entry beats earlier entries *on the same branch*; it cannot see a sibling branch claiming the same number and the same timestamp. The durable rule is the one this violated — a migration's `when` must be a real wall-clock `Date.now()` at authoring time, never a hand-picked offset.

**Merge fallout, taking main's side:** `WidgetFrame` no longer carries a `ConfigEditor` prop (the optional `onRemove` guard is re-applied on top of main's shape), and tab persistence moved to `useTabSessionPersistence`.

Re-verified after the merge: renderer suite **646 files / 7736 passed**, migration + journal + query + ipc suites **25/25**, typecheck (web/node/test), lint, ipc:check, check:architecture, check:contracts, i18n:check, docs:impact --strict, docs:build — all green. Native ABI was flipped to node for the main tests and returned to electron afterwards.

PR #1656, opened 2026-08-20
feat(sidebar): drag the sections into your own order, synced per vault

### LINKED ISSUE #1645: Sidebar section order is hardcoded — let users reorder Collections, Bookmarks, Tags
Parent epic: #1625 (madde **G8**) · Kaynak: Reddit geri bildirimi, 20 Ağu 2026 · Sürüm: `v2026-08-17`

> "I'd also like to be able to order the sidebar to choose the order of collections, bookmarks, canvases, etc."

## Durum — bugün mümkün değil, kod seviyesinde sabit

`apps/desktop/src/renderer/src/components/app-sidebar.tsx` bölümleri **elle yazılmış JSX** olarak, kaynak sırasıyla render ediyor:

```
app-sidebar.tsx:493   {/* SCROLLABLE SECTION - Collections, Bookmarks, Tags … */}
app-sidebar.tsx:568   <NotesTree … />                 (SidebarSection içinde)
app-sidebar.tsx:683   <SidebarSection id="tags"> <SidebarTagList … /> </SidebarSection>
```

Sıra veriden gelmiyor; `sidebar-nav.tsx:29`'daki *"Already filtered to the visible sections, in display order"* yorumu üst nav içindir, bu bölümleri kapsamaz. Bölüm **içinde** sürükleyerek sıralama yalnızca projelerde var (`sortable-project-list.tsx`).

## Yapılacak

- [ ] Bölümleri bir diziden sür: `{ id, label, render }` listesi + kalıcı `sectionOrder: string[]`
- [ ] Bölüm başlıklarına sürükle-bırak (dnd-kit zaten projelerde kullanılıyor)
- [ ] Sıra ayarlarda saklansın ve cihazlar arası sync'lensin
- [ ] **Geriye dönük uyum:** kayıtlı sırada olmayan yeni bir bölüm eklendiğinde varsayılan konumuna düşmeli, listeden kaybolmamalı. Bilinmeyen id'ler sessizce atılmalı.
- [ ] Bölümü gizleme de aynı ekranda mantıklı — kapsamı büyütmeden önce karar ver

Yakın istek: kullanıcı "collections, bookmarks, canvases" diyor; bugünkü bölümlerin tam listesini `app-sidebar.tsx`'ten çıkar ve issue'ya yaz.

h4yfans:
Closes #1645.

## What

The sidebar's five sections (Collections, Projects, Bookmarks, Canvases, Tags) rendered in the order they were written into `app-sidebar.tsx`. Nothing about that order came from data, so nobody could change it. Now each section header carries a drag handle and the order is yours.

- **Handle, not header.** The grip sits in the gutter the header already leaves free, so revealing it on hover shifts nothing, and the header's own click still expands and collapses. Putting dnd-kit's listeners on the header button itself would have taken <kbd>Space</kbd>/<kbd>Enter</kbd> away from toggling; on the grip they give keyboard reordering for free.
- **Persisted + synced.** The order lives in `sidebar.sectionOrder`, next to `sidebar.sortModes` from #1648 — per vault, synced, one field clock for the whole list. Reordering moves every section at once, so two devices dragging concurrently must resolve to one of the two orders; per-entry clocks would merge them into a third order neither user asked for.
- **Forgiving in both directions** (the backward-compat requirement in the issue): ids this build does not render are dropped, and a section missing from a saved order is re-inserted *after the default sibling that precedes it* rather than appended to the bottom. A section added by a later version lands in its intended slot, and Canvases keeps its place across its feature flag going off and back on.
- **No nested DndContext.** The sidebar already sits inside the app-level one (`App.tsx` → `DragProvider`); nesting would re-register the project rows on the inner context, which is what makes dropping a task onto a sidebar project work. The reorder subscribes with `useDndMonitor` instead, and a drop on anything that is not a section leaves the order alone.

Section *hiding* — the last, optional bullet on the issue — is deliberately not in here. It is a separate decision about what a hidden section does to the ⌘-number shortcuts and the drill-down surfaces, and folding it in would have doubled the surface of this change.

## Also

`SidebarSectionAction` extracts the icon-button-with-tooltip that `app-sidebar.tsx` repeated six times. Without it the file crosses the 800-line lint ceiling.

## Tests

- `sidebar-section-order.test.ts` — the resolver: unknown ids, duplicates, a section the saved order never saw (first, middle, several at once), and the drop rules.
- `app-sidebar.section-order.test.tsx` — the sidebar renders the saved order, slots in what the order does not mention, ignores an id it cannot render, and names every handle.
- `settings-handler.test.ts` — an inbound order is written to the data DB and broadcast, and a merge that carries no order leaves the stored one alone.
- `sidebar-section-order.e2e.ts` (5 cases) — the real drag: default order on a fresh vault, drag + persist + survive reload, an older build's saved order, a drag with the sections expanded over project rows, and a drop on itself.

Mutation-checked: breaking the fallback slot fails the unit tests, and a no-op reorder handler fails two e2e cases.

## Verification

`pnpm typecheck` · `pnpm lint` · `pnpm check:architecture` · `pnpm check:contracts` · `pnpm ipc:check` · `i18n:check` · vitest renderer 7814 / main 7182 / shared 2558 / preload 30 · `pnpm test:e2e sidebar-section-order` 5 passed, plus `sidebar-reveal-new-note` + `settings-persistence` as neighbours · `pnpm docs:impact --strict` + `pnpm docs:build`.

Note: `pnpm test` also fails on `packages/i18n`'s `icu-brace-style` for `canvas.link.useUrl` — pre-existing on `main`, untouched here.


h4yfans:
Follow-up: the drag itself was broken, not just rough — fixed in `dc2998408`.

**Why it snapped back.** The sidebar shares the app-level `DndContext`, whose collision detection answers `project` for any drag whose pointer is inside a sidebar project row. dnd-kit nulls a sortable's transform the moment `over` is not one of that sortable's own items (`useSortable`: `displaceItem` needs a valid `overIndex`), so the section jumped back to its resting place mid-drag and a drop there reordered nothing. A section drag now only ever resolves to another section.

**Why it "got lost".** Among sections it resolved by closest *center* — and a section's center is the middle of its content, nowhere near the handle once Collections has its tree open. Sections now resolve by pointer.

**Feedback rebuilt around the same fact.** Sections are as tall as their content, so sliding them out of each other's way threw the sidebar around, and carrying one covered whatever it was being dropped onto. The section now stays put and dims, its name rides the pointer as a small label, and a terracotta line marks the boundary it will land on. Nothing else moves. The label is driven by an offset this component tracks itself: the app mounts a `DragOverlay` for tasks, and a sortable inside a context that has one never receives the drag source's own transform.

**Regression tests** (e2e, 8 cases in `sidebar-section-order.e2e.ts`): the dragged label keeps following the pointer while it is over a project row (this asserted `-3px` before the fix), a tall expanded Collections still reaches the bottom of the list, and the target section shows `data-drop-edge` while every other section's transform stays `0`. Plus unit tests for `sectionDropEdge`.

Issue #1670, opened 2026-08-20
Navigate from a map node back into the note


h4yfans:
## Parent

Part of #1667.

## What to build

The map becomes navigation, not just a picture.

Clicking a node in the map closes the map and lands the user at that place in the note. Two steps are both required: the editor is hidden while the map is showing, and the content area gates its own rendering, so the target block may not exist in the document yet. The existing "scroll to a block that may not be rendered yet" mechanism is what handles that.

The outline panel stays visible in map mode, and clicking a heading there must behave identically — one control must not have two behaviours depending on what else is on screen.

## Acceptance criteria

- [ ] Clicking a heading node closes the map and lands at that heading, including when the target block has not been rendered yet.
- [ ] Activating a node from the keyboard does the same thing as clicking it.
- [ ] Clicking the root node returns to the top of the note.
- [ ] Clicking a heading in the outline panel while the map is showing takes the identical path as clicking the corresponding map node.
- [ ] Navigation works both from the tree-role projection and from the drawn map.
- [ ] Tests cover each path and assert that the outline and the map call the same handler.

## Blocked by

- #1668
- #1669

Issue #1667, opened 2026-08-20
Note mind map: a note's structure as an in-place canvas view


h4yfans:
## Problem Statement

A note's structure is invisible while you are inside it. Headings, nested lists,
tasks and wiki links form a real hierarchy, but the editor shows that hierarchy
one line at a time, and the outline panel flattens it into a 1D list. When a
note grows past a screenful, the user loses the shape of their own thinking:
which sections are heavy, which are stubs, where the branches split, and which
other notes this one reaches out to.

The two surfaces that exist today each answer a different question. The outline
panel answers "where am I right now". The graph view answers "what does this
note link to". Neither answers "what is the shape of this note".

## Solution

A mind map of the note's own structure, rendered in place of the note body
behind a toggle in the note header.

Press the toggle and the title, properties, tags and editor body are replaced by
a mind map: the note title as the root, headings branching by level, list items
and tasks branching underneath, container blocks (toggles, callouts) opening
into their children, and wiki links appearing as their own visually distinct,
clickable leaves. The header, breadcrumb and overflow menu stay exactly where
they were. Press the toggle again and the note comes back, unchanged, with undo
history and cursor position intact.

Every node is clickable: a heading or list node returns to the note and scrolls
to that block; a wiki-link node opens the linked note; a task node opens the
task.

The map is a derived, read-only view — it is regenerated from the note, never
edited in place, and it is never a canvas file the user has to maintain. A
separate action exports it (PNG/SVG) or mints a one-shot canvas snapshot that
then detaches completely and becomes an ordinary, fully user-owned canvas.

## User Stories

1. As a note-taker, I want to see my note's headings as a branching map, so that
   I can grasp the note's shape without scrolling through it.
2. As a note-taker, I want the note's title to be the root of the map, so that
   the map is unmistakably about this note.
3. As a note-taker, I want sub-headings to branch off their parent heading, so
   that I can see how sections nest.
4. As a note-taker, I want list items to branch off the heading they sit under,
   so that I can see the detail beneath each section.
5. As a note-taker, I want nested list items to branch off their parent item, so
   that indentation I already wrote is preserved visually.
6. As a note-taker, I want numbered list items to keep their numbers in the map,
   so that ordered steps stay ordered.
7. As a note-taker, I want toggle blocks to open into their children in the map,
   so that content I collapsed in the editor is still discoverable.
8. As a note-taker, I want callouts to branch into their children, so that a
   callout holding a checklist is not reduced to a single node.
[...]

PR #1686, opened 2026-08-21
test(notes): cover the mind-map journey end to end

h4yfans:
## Why this exists

Epic #1667's testing decisions ask for exactly one end-to-end journey:

> open a note, toggle, see the map, click a heading node, land back in the editor at the right block.

None of the eight tickets' acceptance criteria claimed it — #1670 left it out deliberately, since its own ACs did not list it — so it was the one thing the epic asked for that nothing owned. This is it.

It matters more than its size suggests. The drawing is a bitmap surface: no unit test can see a node inside it, click one, or read a label off it. Everything below the canvas is proven by the pure entry point and by renderer tests against the accessible tree; this is the only test that proves the whole thing works when it is actually running.

## What it covers

**1. The journey.** Seed a note with five headings and enough filler to make the page taller than the viewport, open it, press the header toggle, and assert:

- the toggle reports `aria-pressed="true"` and the map appears;
- the editor is **attached but not visible** — the two halves are separate assertions on purpose, because "attached" is the undo-history guarantee and "not visible" is the replacement;
- the map region's label names the note;
- all five headings reached the tree;
- activating the **last** heading closes the map, brings the editor back, and lands on that heading — and the note's first `h1` is no longer in the viewport, which is the assertion that fails if the toggle closes without navigating.

The note is long on purpose. In a short note "the heading is on screen" is already true before the click, and the test would pass without the feature working.

**2. The undo guarantee.** The unit suite asserts it structurally — the editor instance is the same object across a toggle round trip. This asserts what that buys: type, look at the map, come back, undo, and find the typing undone rather than the history gone. Had the map unmounted the editor, the collaborative undo manager would have gone with it.

## Keyboard, not pointer — deliberately

The node is activated with `Enter`, not a click. The tree is `sr-only`, one pixel underneath the drawing, and a pointer never reaches it in the product either: a mouse user clicks the drawn box on the canvas, a keyboard or screen-reader user walks this tree. Driving it from the keyboard is what the layer is for, and it doubles as the acceptance check that keyboard activation does the same thing a click does.

Two earlier failures were both the test being wrong, not the feature: driving an `sr-only` layer with a pointer, and a block id that appears on more than one node in a rendered block (the app's own `scrollToHeadingBlock` scopes its lookup and takes the first match for the same reason).

## Verification

Run against a fresh production build — the suite launches `out/` and only rebuilds when `BUILD_BEFORE_TEST` is set, so a stale bundle silently tests old code:

```
MEMRY_ENV=production pnpm exec electron-vite build     exit 0   ✓ built in 13.03s
bash scripts/ensure-native.sh electron                 exit 0

pnpm exec playwright test tests/e2e/mind-map.e2e.ts    exit 0
  ✓ 1 [electron] › Note mind map › opens a note as a map and lands back on the heading a node names (14.0s)
  ✓ 2 [electron] › Note mind map › keeps undo history across a trip to the map and back (13.9s)
  2 passed (28.3s)
```

Repository gates:

```
pnpm lint                                    exit 0   108 problems (0 errors, 108 warnings), all pre-existing
pnpm typecheck                               exit 0   Tasks: 17 successful, 17 total
pnpm --filter @memry/desktop i18n:check      exit 0   ok: i18n check passed
pnpm docs:impact --base origin/main --strict exit 0   no desktop/sync-server docs-relevant changes detected
```

One new test file, no production code touched.

Part of #1667.

### DIFF
--- a/apps/desktop/tests/e2e/mind-map.e2e.ts
+++ b/apps/desktop/tests/e2e/mind-map.e2e.ts
@@ -1,0 +1,162 @@
+/**
+ * Note mind map E2E
+ *
+ * The one end-to-end journey the mind-map epic (#1667) asks for: open a note,
+ * press the header toggle, see the note's structure as a map, click a heading
+ * node, and land back in the editor at that heading.
+ *
+ * This journey is only reachable because the map ships a `role="tree"` DOM
+ * projection alongside the drawing. The drawing itself is a bitmap surface —
+ * Playwright cannot see a node inside it, cannot click one, and cannot read a
+ * label off it. The tree layer exists for screen readers, for keyboard users,
+ * and for exactly this file.
+ *
+ * The second test is the undo guarantee. The unit suite asserts it structurally
+ * (the editor instance is the same object across a toggle round trip); here it
+ * is asserted the way a user would notice it — type, look at the map, come back,
+ * undo, and find the typing undone rather than the history gone.
+ */
+
+import { test, expect } from './fixtures'
+import type { Page } from '@playwright/test'
+import {
+  waitForAppReady,
+  waitForVaultReady,
+  seedNote,
+  SELECTORS,
+  SHORTCUTS
+} from './utils/electron-helpers'
+import { waitForNoteById, openNoteByHandle } from './utils/note-sync-helpers'
+
+const NOTE_TITLE = 'Mind Map Journey'
+
+/**
+ * Long on purpose. The point of the last assertion is that the editor SCROLLED
+ * to the clicked heading, so the note has to be taller than the viewport —
+ * otherwise "the heading is on screen" is true before the click as well and the
+ * test passes without the feature working.
+ */
+const FILLER = Array.from(
+  { length: 12 },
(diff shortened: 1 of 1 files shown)


h4yfans:
Completing the gate output — the repo-wide suite finished after the PR was opened.

```
pnpm test                     exit 1
  [ERR_PNPM_RECURSIVE_EXEC_FIRST_FAIL] Command was killed with SIGSEGV
  (Segmentation fault): vitest run --config config/vitest.config.ts
  Tasks: 8 successful, 9 total
  No failing spec named, no "Test Files ... failed" line — the worker was killed.

pnpm test:desktop             exit 0
  Test Files  1379 passed | 3 skipped (1382)
       Tests  17991 passed | 1 expected fail | 13 skipped (18005)
```

The `pnpm test` red is the known turbo flake (the runner is killed mid-stream, no test is reported failing); the direct desktop run is green. Reported rather than summarised, since a red gate that reads green is exactly the failure mode worth avoiding here.

PR #1704, opened 2026-08-22
fix(editor): a wiki link click navigates again, and a cell takes pictures and checkboxes

h4yfans:
Three things a user reported after the last release.

## A wiki link click stopped navigating

> "They only seem to work when viewing the Mind Map Mode... I have an index page with a list of links and when I click one it goes into edit mode rather than to the linked page."

The chip is an atom rendered `contenteditable="false"`, so mousedown parks the caret beside it, and the edit plugin then paints the raw `[[Target]]` — hiding the chip with `display: none`. Chromium flushes that repaint off `selectionchange` **between mousedown and mouseup** on a real ~100 ms click. By mouseup the chip element is gone, the browser retargets `click` to the paragraph, and the DOM listener's `closest('[data-wiki-link]')` finds nothing. Navigation dies silently and the user is left looking at painted markdown — which reads exactly as "it went into edit mode". Regressed in 52c6cd07f.

It is a race, not a link count: a fast enough click still lands on the chip, which is why a single link "worked fine" and why the existing e2e — Playwright's zero-delay `.click()` — stayed green through the whole regression.

Navigation now runs from ProseMirror's `handleClickOn`, off the position captured at **mousedown**, which no repaint can move. Returning true also suppresses the caret placement, so the paint never appears at all. The DOM branch is removed, so there is exactly one navigation path and a fast click cannot open the note twice.

Also fixed alongside: `raw: text` in the widget spec. `raw` is a ProseMirror boolean, not a payload — passing a string made the painted `[[…]]` editable DOM that is not in the document, so the caret could enter it and typing went nowhere.

## An image pasted into a table cell did nothing

The interceptor only handled clipboard payloads carrying a real `File`. Copying a picture from another note, a web page or a doc delivers `text/html` with no file, which fell through to BlockNote, was parsed as an image **block**, and was silently dropped by a cell that holds inline content only.

Paste and drop now read the HTML fragment and insert `inlineImage`. Remote and `data:` sources are accepted — base64 is saved as an attachment rather than written into the markdown row. A ref belonging to *another* note is deliberately not taken: its path is relative to that note and would arrive broken. A fragment mixing a picture with text is left to the ordinary paste rather than half-handled.

## Checkboxes could not exist in a table cell

`checkListItem` is a block, and no position inside a cell can hold one — so BlockNote's `[ ] ` input rule, its `Mod-Shift-9` shortcut and its slash item all bail out. Adds an `inlineCheckbox` inline spec on the same pattern `inlineImage` established: typing `[ ] ` in a cell makes a tickable box, `/check` inside a cell puts it in the cell, and a cell someone wrote by hand as `| [x] task |` opens as a real checkbox.

## On disk

No new format. A cell checkbox serializes to the plain text a markdown table already carries — `| [x] task |` — so Obsidian and every other markdown editor read the row unchanged, and the bytes round-trip byte-stable. Cell images were already `alt`.

**Compat.** An older client sharing a Y.Doc does not know `inlineCheckbox` and will drop it. The node is an atom, so the label text survives and only the marker is lost — strictly smaller than the `inlineImage` precedent, where the whole picture reference goes. Attributes arriving from a synced doc are strings, so `Boolean("false")` would tick every box on the second device; that is guarded and tested both in the package and against the real converter.

## Verification

- `wiki-link-open-promotion.e2e.ts` — new spec performs a **human-shaped** click (mouse.down → 150 ms → mouse.up) on the second link of an index note. Confirmed **red before, green after**.
- `table-cell-checkbox.e2e.ts` — new file, 3 specs. Confirmed **red before (3 failed), green after**.
- `table-cell-image.e2e.ts` — new HTML-clipboard paste spec.
- All 13 specs across the three files: **green serially**. (Under 2 parallel workers, 2 of them flake with "browser has been closed" — resource contention, green in isolation and green at `--workers=1`.)
- `pnpm typecheck`, `pnpm check:architecture`, `pnpm check:contracts`, `pnpm docs:impact --strict`, `pnpm docs:build`: all exit 0. Lint: 0 errors, and none of the 107 pre-existing warnings are in touched files.
- Renderer suite green apart from `notes-tree-folder-rename-duplicate.test.tsx`, a 30 s timeout under full-suite CPU load that passes 14/14 standalone and imports nothing touched here.

The e2e specs for table cells run in **no CI job** (`grep table-cell .github/workflows/` is empty) — worth wiring up separately.

## Known follow-ups, deliberately not in this PR

- A paste that mixes text and a picture still loses the picture in a cell.
- Copying an image out of note A into a table in note B still does nothing; the fix needs a resolve-then-fetch against the source note.
- The toolbar list toggle and `Mod-Shift-9` inside a cell are still no-ops (BlockNote's own `content !== 'inline'` guard).
- Cell checkboxes are not tasks — `scanTaskCheckboxStates` requires a line-leading list marker, so they carry no `{task:id}` and never reach the task list.
- Checkbox promotion of an externally-edited cell runs on open only, not mid-session.


h4yfans:
### Follow-up: the click no longer flashes `[[Target]]` first

Reported after the first round: the click navigated, but the link briefly rendered as raw `[[abc]]` — and with "clicking a page opens a new tab" on, that flash lasted until the new tab appeared.

`handleClickOn` runs at **mouseup**, and the caret is placed natively at **mousedown**, so the paint had already happened and `preventDefault()` on the mouseup could not undo it. The press is now claimed at mousedown in `handleDOMEvents`: `runCustomHandler` returns `handler(view, event) || event.defaultPrevented`, so preventing the default makes ProseMirror skip its own mousedown entirely — no `MouseDown`, no caret placement, no `selectionchange`, so `decorations()` never recomputes and the chip is never hidden. Navigation moved to the matching `mouseup`, guarded by the same 4px slop PM uses, so a drag that started on a chip is not a click.

**Also fixes a regression this PR introduced.** The first round's `handleClickOn` bailed on `shift|alt|meta|ctrl`, and the DOM listener that used to catch those was removed in the same change — so ⌘+click and ⌥+click on a link did nothing at all. They navigate again; only shift stays excluded so a selection can still be extended across a link.

**Tradeoff:** with PM's mousedown skipped for chip presses, a text-selection drag that *starts on a chip*, and double/triple-click starting on a chip, no longer expand. Both have an obvious substitute one character over. Documented in the user guide.

**Docs correction.** `wiki-links.md` claimed "⌘+click to open in the background; ⌥+click to open in a split pane". That was never true for wiki links — it was written in a docs-only commit (dd57af457) and no code in this path has ever read a modifier; `onNavigate` receives a bare string and `openLinked`/`openTab` only consult the reuse preference. Background tabs and split panes do exist (`OpenTabOptions.background`, `openInNewTab`/`openToTheSide`) but are wired to sidebar rows only. The line now describes what actually happens. Wiring the modifiers up for real is a separate, small piece of work if we want it.

Same class of unverified claim, flagged but untouched: `apps/docs/src/user-guide/search.md:76` ("⌘+⌥+Enter — open in split pane") — no `altKey` anywhere under `components/search/`.

**Verification.** The e2e now reads the editor *while the button is held* and asserts no `[data-wiki-link-source]` widget is on screen. Red before (`Expected: 0, Received: 1` — the flash, reproduced), green after; all 5 specs in the file pass. Renderer suite 683 files / 8310 tests green, `typecheck:web`, `typecheck:test` and lint clean.

h4yfans:
Rebased onto current main (merge commit, clean). Local verification on the merged head:

- `wiki-link-edit-plugin.test.ts` + `use-block-note-setup.test.ts`: 58/58
- `@memry/editor-schema`: 120/120
- `pnpm typecheck`: green (all packages)
- `wiki-link-open-promotion.e2e.ts` on a real production build: **5/5**, including the human-held click spec (`mouse.down()` → 150ms → `mouse.up()`) that asserts no `[[…]]` paint while the button is down, the chip stays visible, and navigation lands.

Context: the single-click regression users still see in 2026.823.x is exactly the pre-fix behavior — this PR never merged, so the fix never shipped. The only red on the previous run was the known `fts-corruption` coverage flake.

Note for merge ordering: #1717 branched from main without this PR and still carries the old DOM click listener in `use-block-note-setup.ts`. Merge this first, then rebase #1717 — its confirm-create dialog hangs off `onInternalLinkClick`, which this PR still routes through `onNavigate`, so it composes cleanly after rebase.

PR #1706, opened 2026-08-23
feat(sync): per-platform write gate and the @memry/sync-client scaffold

h4yfans:
## Summary

Spec `001-mobile-app` **Phase 2 (Train Phase 1)** — the server production-safety kit lands complete, `@memry/sync-client` is scaffolded with its ten platform seams, and **the extraction is now half done: 54 of ~105 platform-free sync files have moved out of desktop**, with desktop green after every slice. **Stacked on #1705**; review that first.

**G1 is not green yet.** T018 is in progress (54 files moved), T019–T022/T024/T025/T032 open. The seam question that was blocking the extraction is decided in this PR (below).

### Seam amendment — decided, in this PR

The T015 seam inventory found that **vault file I/O (~20 files) mapped to none of the ten seams**: `VaultDirectory` owned vault roots, `AttachmentStore` is keyed by attachment id, `CrdtStorePath` is the CRDT store's location — and nothing owned note/journal reads and writes. `note-handler.ts` alone does `mkdirSync` / `writeFileSync` / `renameSync` / `readdir` / `rmdir` on note paths; there was no legal platform-free way to say that.

**Decision (owner, 2026-08-23): widen seam 6 into `VaultFileSystem`** — roots *plus* relative-path `readFile`/`writeFile`/`readBytes`/`writeBytes`/`exists`/`remove`/`rename`/`list`/`removeDirIfEmpty` — rather than add an eleventh seam. Roots and contents are the same concern at the same boundary, so **the seam count is still ten**. Recorded in `contracts/platform-adapters.md` §6 and in the inventory; the inventory now has zero unassigned files.

The content half deliberately matches the existing [`NoteContentStore`](packages/storage-vault/src/note-content-store.ts), which desktop already implements over `node:fs` and which T034 already commits mobile to — so this formalises a contract both shells were heading for anyway. Design points worth reviewing: paths are always relative to the vault root (absolute paths are the one thing that cannot mean the same thing on both shells); writes are atomic and create their own parents, so there is deliberately **no `mkdir`**; and `removeDirIfEmpty` takes an explicit ignore list rather than deleting recursively, because the recursive version of that call is a data-loss bug.

### Server production-safety kit — T027–T031, complete

- `0006_client_gate.sql`: `client_policies` table plus nullable `client_platform` / `client_version` on `sync_items`, `crdt_updates`, `crdt_snapshots`, and a partial index for the incident query. **The table is empty and the columns are NULL after this migration**, which is exactly today's behaviour; nothing is seeded and nothing is backfilled. No `desktop` row is created on purpose — that would be a live kill switch pointed at the only shell that exists.
- `client-gate` middleware parses `x-memry-client` and gates **writes only**; reads are never gated, so a device dropped to read-only still opens every note it owns. Absent *or malformed* header means legacy desktop and full access — a header-parser bug must not be able to lock a paying user out. Pre-release versions (`1.0.0-beta.1`) count as malformed, so a beta can never satisfy a floor its release does not.
- Kill switch is evaluated **before** the version floor: with writes off for a platform, `CLIENT_UPGRADE_REQUIRED` would send users chasing a release that cannot help them. Every uninterpretable policy — no row, NULL floor, unparseable floor — resolves to **allow**, so an unreadable policy table degrades to current behaviour rather than to an outage.
- Writes are stamped with the calling platform and semver. Attribution records the **latest** writer, not the creator. `crdt_updates` / `crdt_snapshots` are included deliberately: a note's body lives there, and that is the payload most likely to need a targeted rollback after a mobile incident.
- `GET /sync/status` echoes the caller's own policy as an optional `clientPolicy`, so a device learns about a flipped switch on its next foreground poll instead of by attempting a write and eating a 403. Header-less clients get byte-identical status responses and pay no extra query.

### `@memry/sync-client` — T015–T017, T023

- [`docs/seam-inventory.md`](packages/sync-client/docs/seam-inventory.md) + a regenerable scanner: 174 non-test files across `apps/desktop/src/main/sync`, `packages/app-core`, `packages/storage-vault`; 69 platform-touching, all assigned.
- The ten adapter interfaces, one file per seam, lifted from `contracts/platform-adapters.md`. The tsconfig sets `types: []` so node globals cannot leak into platform-free code.
- One **runner-agnostic** adapter conformance suite (`describe`/`it`/`expect` injected) so desktop's Vitest run and mobile's on-device run share one suite instead of drifting into two.

### Extraction — T018 in progress, 54 of ~105 files moved

Three slices, one commit each, desktop green after each:

1. **`vector-clock`** — the mechanism proof: workspace wiring, subpath export, 26 call sites repointed.
2. **15 zero-dependency modules + the Drizzle type widening.** compress, field-merge, crdt-payload, crdt-preflight-protocol, crdt-snapshot-watermark, note-events, note-size, worker-protocol, agent-conversation-fields, microtask-batch-broadcaster, reminder-outbound, settings-sync-keys, sync-eligibility, sync-timer, task-activity-retention move with their 10 platform-clean suites. The T015 finding lands: canonical `DrizzleDb` / `IndexDrizzleDb` live in `@memry/db-schema/drizzle-db` as `BaseSQLiteDatabase<'sync', { changes: number }, schema>` — the structural intersection of better-sqlite3's and expo-sqlite's run results, both drivers synchronous, so **no call site gains an `await`**. All 23 type-only-bound files, `SyncQueueManager`, `storage-data`'s `NoteMetadataDb` and the query helpers the type flows through are now platform-free.
3. **The outbox spine + record services + platform-free handlers, behind two new facades.** `sync-client/src/logging.ts` (lazy scoped logger, factory-injected) and `src/telemetry.ts` (event/log sink-injected) let extracted modules keep byte-identical logging/telemetry call sites; desktop wires both in `main/sync/sync-client-runtime.ts`, the first import of the main entry. Moved: `queue`, `offline-clock`, `current-device-id`, `attachment-download-state`, `concurrency`, `crdt-compact-utils`, `crdt-snapshot-scheduler`, all 19 platform-free record `*-sync` services, `item-handlers/types` + 11 platform-free handlers. 131 desktop call sites repointed, `vi.mock` specifiers included.

**Test policy through the move (T022):** DB-backed suites stay in desktop on real better-sqlite3 fixtures and import the package; only platform-clean suites move (12 so far). Three test-only adjustments, none touching an assertion, each recorded in its commit: two suites mocked `microtask-batch-broadcaster` by old relative path and went **inert** once the provider imported it from the package (one failed honestly, one was a silent false-green) — `vi.mock` specifiers moved to the package path; `offline-clock.test`'s logger mock re-targets `@memry/sync-client/logging` for the same reason; one moved suite needed two type-only fixes to leave desktop's typecheck **exclude backlog** (which shrinks by one).

**Still in desktop, with the reason:** `retry` (imports http-client's error classes — extract those first), `engine`/`network`/`websocket`/`attachment-events` (`node:events` → platform-free emitter, planned), `blocknote-converter`/`agent-message-handler` (`node:crypto` → WebCrypto/`@noble/hashes`), `certificate-pins`/`sync-server-url`/`cert-hash-cli` (`process.env` under a `types: []` package), and everything gated on the VaultFileSystem seam (T021), the CRDT seams (T019) and the remaining cross-cutting facades (db accessor, i18n, store, window-broadcast).

### Findings worth knowing

- **The seam inventory undercounts the real coupling.** Scanning for node builtins and platform packages finds 69 files, but the extraction's actual blockers are desktop-internal cross-cutting imports — `lib/logger` alone was an edge into ~100 files, `database/index`, `telemetry/*`, `main-i18n`, `store`, `window-broadcast`, the vault layer. The facades + type widening in this PR clear the first three; the rest map to T019–T021.
- **One sync-hash to substitute.** `item-handlers/agent-message-handler.ts` calls `createHash('sha256')` inside `applyUpsert`, which `SyncItemHandler` declares as synchronous. WebCrypto's `subtle.digest` is async, so the fix is `@noble/hashes` (already in the tree, already a direct dep of `apps/mobile`) rather than making the whole handler interface async. Lands with the next T018 slice.
- **A `;` inside a *trailing* `--` comment in a D1 migration breaks the test harness.** `tests/sync-harness` drops only full-line comments and then splits on `;`, so it tore `CREATE TABLE client_policies` in two and three desktop integration files died with `D1_ERROR: incomplete input` — while `pnpm --filter @memry/sync-server test` stayed green, because wrangler and better-sqlite3 parse SQL properly. Rule recorded in `apps/sync-server/migrations/README.md`.

### Deviation from the contract sketch

`contracts/sync-protocol-additions.md` §2 sketched a flat `{ code, minVersion }` rejection body. Implemented as the project's existing envelope instead — `{ error: { code, message, minVersion? } }` — because every shipped client already reads `error.code`. Recorded in the contract doc.

## Compatibility

Additive throughout: new table (empty), new columns (nullable, no backfill, no read path depends on them), new optional request header, new optional response field. No existing endpoint changes shape or status code for a header-less client. A Worker deployed before this migration, or rolled back after it, never touches the new columns.

The extraction is behaviour-preserving by construction: moved files are byte-identical modulo import specifiers, the type widening adds no `await` and no runtime change, and the facades forward to the exact functions the moved code called before. `legacy-client-compat.test.ts` still proves the server side against a **real** SQLite D1.

## Verification

| Check | Result |
|---|---|
| `pnpm lint` | green |
| `pnpm typecheck` | 18/18 |
| `pnpm test:desktop` | 1380 files (1377 passed, 3 skipped) / 18033 tests (18019 passed, 13 skipped, 1 expected fail), 0 failures |
| `pnpm --filter @memry/sync-client test` | 13 files / 117 tests |
| `pnpm --filter @memry/sync-server test` | 66 files / 1032 tests, coverage thresholds met |
| `pnpm check:architecture` · `pnpm check:contracts` · `pnpm ipc:check` | green |
| `pnpm docs:impact --base origin/mobile-phase0-g0-spike --strict` · `pnpm docs:build` | green |

Docs: `apps/docs/src/architecture/sync-protocol.md` gains a "Client Identification and the Per-Platform Write Gate" section plus two error-mode rows.

## Still open in Phase 2

Closed by the stacked follow-up **#1719**: T019 (CRDT merge/pending behind the seams), T020 (all ten desktop adapters, conformance-proven), T021 (app-core split — node cluster to `apps/cli`, package fully platform-free), T022, T024 (real mobile edge + red/green boundary re-proof), T025 (targeted E2E 12/12), T026, T032 (`@memry/mobile` in the root filters), and the T018 tail (http errors, emitter, crypto/env substitutions).

Open after #1719, on record: T018's engine knot — 60 platform-free files that move only when the engine consumes the seams instead of the concrete modules (behaviour-affecting refactor, split out; see #1719 and `seam-inventory.md` §Surface).

Note on T032: the root `typecheck` / `test` filters are explicit **allowlists**, not exclusions, so it was an add rather than a removal.


h4yfans:
Landed on main via #1721's merge commit (the full Phase 0→G1→US1 lineage; this branch's commits are reachable from main). Closing rather than merging because GitHub does not retroactively mark stacked PRs merged after a retarget.

Issue #1713, opened 2026-08-23
Note attachments panel + self-heal for broken attachment refs


h4yfans:
Part of #1708 (retrievability epic). Optional hardening — lowest priority in the epic.

## Problem

Attachments are second-class for retrieval:

- There is no UI listing a note's attachments; the `listAttachments` / `deleteAttachment` IPC exist (`packages/rpc/src/notes.ts:597-623`) but have **zero renderer call sites**.
- `attachments/` is excluded from the indexer and the file watcher (`apps/desktop/src/main/vault/indexer.ts:330`, `watcher.ts:228`), so an externally renamed/moved attachment breaks its block reference **silently** and nothing ever repairs it — the PDF block just shows the "Failed to load" card forever.

## Work

1. **Attachments panel:** a "Attachments" entry in the note menu (near Reveal in Finder / Copy path in `pages/note.tsx`) listing the note's files via `listAttachments` — original name (from block props where available), stored name, size — with per-row Reveal in Finder and Open actions (reuses the IPC from the file-block issue in this epic).
2. **Self-heal pass:** when a file block's `url` fails to resolve, attempt a repair within the same `attachments/<noteId>/` folder by matching the name suffix after the 6-char nanoid prefix (`{prefix}-{name}` scheme, `apps/desktop/src/main/vault/attachments.ts:313-328`); on a unique match, rewrite the block's `url` through the normal editor update path. Never guess across notes; ambiguous matches stay broken but should surface the "Failed to load" card with the expected filename so the user can fix it manually.

## Acceptance

- User can see and reveal every attachment of a note from the note menu.
- A prefix-preserving external rename no longer permanently breaks the block, and an unrepairable block tells the user which filename it expected.

PR #1812, opened 2026-08-24
file/image blocks: rename an attachment from the block menu

### LINKED ISSUE #1714: file/image blocks: rename attachment from the block menu
Split out of #1709 (deliberately out of scope there). Part of the retrievability epic #1708.

## Problem

There is no UI to rename an attachment anywhere in the app. The on-disk name is `{6-char-nanoid}-{sanitized-original-name}.ext` under `attachments/<noteId>/`, and the original filename lives only in the block's `name` prop / `<!-- file:{…} -->` marker. Users who attach a badly named file ("scan_0031.pdf") can never fix it.

## Work

- Add **Rename…** to the file/image block menu introduced in #1709.
- Renaming must update, atomically enough to survive a crash mid-way: the block `name` prop + serialized marker, the on-disk filename (keep the nanoid prefix), and any references to the old relative path.
- Define sync semantics: the rename must propagate as a note body change + attachment path change without re-uploading the blob, and older clients must tolerate the new path.
- Decide collision behavior when the target name already exists in the note's attachment folder.

## Acceptance

Renaming a PDF from the block updates the visible name, the stored file on disk, and syncs to a second device without breaking the embed.

h4yfans:
## Summary

Adds **Rename…** to the attachment block menu (file / PDF / audio cards, image blocks, and the right-click context menu), and makes that rename land on every device's disk.

Closes #1714.

**What a rename does**
- Renames the file on disk, keeping the `{6-char nanoid}-` prefix and the extension. The dialog edits only the middle segment; the extension is shown beside the field, not edited.
- Rewrites the block's `url` + `name`, which is what records the rename in the note.
- Collisions inside the note's attachments folder get `-2`, `-3`, … rather than overwriting. Submitting the name the file already has is a no-op.
- Order is **disk first, body second**: a crash in that window leaves the body on the old name, which self-heal (#1713) already repairs by prefix. The reverse order would leave the body naming a file that exists nowhere.

**Sync semantics — nothing is re-uploaded**
The blob and its encrypted manifest are untouched (the manifest's filename stays frozen at whatever it was at upload). Only the note body travels, through the ordinary CRDT path. No new sync item type, no new DB column, no change to the note payload shape — so an older client is unaffected: it sees a body whose ref differs from its disk and falls back to self-heal, exactly as it does today for an externally renamed file.

Peers converge their own disk from two narrow, evidence-backed points:
1. **Write-back diff** (`crdt-writeback.ts`) — when a body replaces another, a ref that disappeared plus a ref that arrived under the same nanoid prefix *is* the rename that just arrived. That device renames its own copy.
2. **Post-download** (`sync-attachment-handlers.ts`) — a file materialized after the rename lands under the manifest's old name, so it is renamed to what the body asks for right after the download.

**Deliberately not a general "make the folder match the body" pass.** A file renamed by hand in Finder (#1713's case) produces no body change and is never touched — it keeps the user's name and keeps being served through self-heal. Only an in-app rename moves a file on disk, and only an in-app rename travels.

## Release note

Attachments can now be renamed from the block menu. The file is renamed on disk — keeping its extension — and the new name reaches your other devices without re-uploading anything.

## Test plan

New tests
- `src/main/vault/attachment-rename.test.ts` (15) — name shaping, extension/prefix retention, collision walk, same-name no-op, "this note's folder only" guard, renaming a self-healed file, empty-name rejection.
- `src/main/vault/attachment-rename-reconcile.test.ts` (17) — ref extraction across file markers / image links / legacy `memry-file://` urls, rename pairing (including two renames in one body change and the ambiguous cases it refuses), the peer-side disk rename, no-op on the originating device, never overwriting an existing target, **and that an external rename with an unchanged body is left alone**, plus the post-download rename.
- `crdt-writeback.test.ts` (+2) — the write-back hands old and new body to the reconcile; a throwing reconcile does not make a successful write-back look failed.
- `sync-attachment-handlers.test.ts` (+2) — an embedded (`intoDir`) download is reconciled; a binary-note download is not.
- `attachment-block-menu.test.tsx` (+3) — Rename is offered only where the result can be written back, the IPC round trip hands the new url/name to the block, a failed rename keeps the dialog open and leaves the block alone.
- `file-block.test.tsx` (+1) — a completed rename is recorded on the block props.
- `attachment-block-menu.e2e.ts` (+1) — rename from the menu → the file is renamed on disk and the note's markdown carries the new ref.

Gates run locally (after rebase onto `main`)
- `pnpm typecheck` ✅
- `pnpm lint` ✅ (0 errors)
- `pnpm --filter @memry/desktop i18n:check` ✅
- `pnpm ipc:check` ✅ (bindings regenerated via `pnpm ipc:generate`)
- `pnpm docs:impact --base origin/main --strict` ✅ · `pnpm docs:build` ✅
- Affected suites: main 91 ✅ (4 files), renderer 25 ✅ (2 files)

Not run locally: the full `pnpm test:desktop` sweep and the new E2E spec — CI covers both.

## Reviewer notes
- `apps/docs/src/user-guide/notes/attachments.md` documents the rename, what stays fixed (extension + prefix), the collision rule, the cross-device behaviour, and explicitly that a Finder rename is never undone.
- The reconcile deliberately refuses to guess: an ambiguous prefix (two candidates on either side of the diff) is left alone, and a prefix-less legacy filename carries no identity to pair on, so it is skipped.

### DIFF
--- a/apps/desktop/src/main/ipc/generated-ipc-invoke-map.ts
+++ b/apps/desktop/src/main/ipc/generated-ipc-invoke-map.ts
@@ -198,6 +198,7 @@ export interface MainIpcInvokeHandlers {
   "notes:add-status-option": (...args: [{ propertyName: string; categoryKey: "todo" | "in_progress" | "done"; option: { value: string; color: string; }; }]) => Awaited<Promise<{ success: boolean; }>>
   "notes:apply-template": (...args: [{ noteId: string; templateId: string; mode: "full" | "body"; }]) => Awaited<Promise<{ success: true; note: import("../vault/notes-crud").Note; }> | { success: false; error: string }>
   "notes:attachment-open-external": (...args: [{ noteId: string; url: string; }]) => Awaited<Promise<void>>
+  "notes:attachment-rename": (...args: [{ noteId: string; url: string; newName: string; }]) => Awaited<Promise<import("../../../../../packages/contracts/src/notes-api").AttachmentRenameResult>>
   "notes:attachment-resolve": (...args: [{ noteId: string; url: string; }]) => Awaited<Promise<import("../../../../../packages/contracts/src/notes-api").AttachmentResolveResult>>
   "notes:attachment-reveal-in-finder": (...args: [{ noteId: string; url: string; }]) => Awaited<Promise<void>>
   "notes:create": (...args: [{ title: string; content?: string | undefined; folder?: string | undefined; tags?: string[] | undefined; template?: string | undefined; }]) => Awaited<Promise<{ success: true; note: import("../vault/notes-crud").Note; }> | { success: false; error: string }>
--- a/apps/desktop/src/main/ipc/notes-handlers.ts
+++ b/apps/desktop/src/main/ipc/notes-handlers.ts
@@ -21,13 +21,15 @@ import {
   ApplyTemplateSchema,
   LargeFileReadLinesSchema,
   LargeFileSearchSchema,
-  AttachmentActionSchema
+  AttachmentActionSchema,
+  AttachmentRenameSchema
 } from '@memry/contracts/notes-api'
 import {
   resolveAttachment,
   revealAttachmentInFinder,
   openAttachmentExternal
 } from '../vault/attachment-actions'
+import { renameAttachment } from '../vault/attachment-rename'
 import {
   openLargeFileSession,
   readLargeFileLines,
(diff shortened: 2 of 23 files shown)


h4yfans:
Follow-up fix pushed: **Rename on an image block did nothing.**

Root cause: `ImageAttachmentMenu` and `ImageHoverMenuButton` are rendered by `ContentArea` only while their menu is open / the pointer is on the image. The rename dialog was a child of those components, so clicking Rename closed the menu, `onClose`/`handleMenuOpenChange` cleared the target, and the dialog unmounted in the same tick it was asked to open. The file-block path was unaffected — its menu button lives in a card that stays mounted, which is why the unit tests and the E2E spec were green.

Fix: the rename flow (IPC call + dialog) is now its own component, `AttachmentRenameFlow`, owned by whoever can keep it mounted. The image surfaces take `onRequestRename` and only ask; `ContentArea` holds the target and renders the dialog. The file block keeps the menu-owned dialog.

Also in this commit: an image block with no display name falls back to the *stored* filename, so submitting the field untouched produced `k3f9x2-k3f9x2-photo.png` (main re-applies the prefix). The field now strips a stored prefix for display.

Tests: `attachment-block-menu.test.tsx` gains a harness that unmounts the menu exactly the way the editor does and asserts the dialog survives and completes the rename, plus a unit test for the prefix-stripping. Renderer 13 + 14 ✅, `pnpm typecheck` ✅, `pnpm lint` ✅.

Pushed with `MEMRY_DOCS_IMPACT_SKIP=1`: this is a bug fix inside a feature the docs already describe — `apps/docs/src/user-guide/notes/attachments.md` already documents Rename… on every attachment block including images, and nothing about the documented behaviour changed.

PR #1827, opened 2026-08-24
docs(mobile): G2 evidence — T054/T055 proven, T056 3/5 drills green

h4yfans:
## What

- **T054 → [x]**: device proof (iPhone 12 Pro: conformance 26/26 + round-trip 6/6, 2026-08-24 21:07) + simulator closure run after the full fix chain (22:50 — bulk ladder applied, nothing stuck, list complete).
- **T055 → [x]**: Maestro `us1-unlock-browse.yaml` green on the iPhone 17 Pro simulator (2026-08-25 01:44) — stored session short-circuits sign-in/unlock as designed, browse→open-note asserted. One flow fix: the note body sets `accessibilityLabel="Note content"` (no testID), so the final assert matches the label instead of `id:`.
- **T056 → 3/5 annotated**: kill-switch ✓ (staging `client_policies` flip → Read-only banner on next bg→fg, no restart; clears on re-enable), version-gate ✓ (99.0.0 floor → update banner → clears on NULL), attribution ✓ (staging D1: zero `ios` write rows across sync_items/crdt_* — US1 pull-only proven).
- G2 checkpoint block refreshed: remaining items are the final phone pass after #1826, the <5 s visibility drill (owner call on trigger path), and SC-004 (owner call on seeding the 10k vault).

## Verification

Staging `client_policies` restored to `writes_enabled=1, min_write_version=NULL` after the drills (verified by query). Screenshots + attribution JSON captured in the session.

### DIFF
--- a/apps/mobile/.maestro/us1-unlock-browse.yaml
+++ b/apps/mobile/.maestro/us1-unlock-browse.yaml
@@ -41,5 +41,5 @@ appId: com.memry.mobile
 - tapOn:
     below: 'Notes'
     index: 0
-- assertVisible:
-    id: 'Note content'
+# label match — the note body sets accessibilityLabel, not testID
+- assertVisible: 'Note content'
--- a/apps/mobile/src/sync/background.ts
+++ b/apps/mobile/src/sync/background.ts
@@ -51,18 +51,45 @@ export async function registerBackgroundSync(minIntervalMinutes = 15): Promise<v
 }
 
 let foregroundWired = false
+let pollTimer: ReturnType<typeof setInterval> | null = null
 
-/** Foreground transitions trigger an incremental sync pass. */
+/** While the app is open, poll every 5 s so desktop edits show without a
+ * background/foreground cycle (G2 <5 s visibility). The engine's exclusive
+ * queue coalesces overlapping passes, and a pass with nothing new is one
+ * cursor-echo request. */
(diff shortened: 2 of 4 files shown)


h4yfans:
Two mobile sync fixes landed on this branch during the T056 drills (both proven on simulator against staging):

1. **Body edits were invisible to mobile** — the record feed never carries body changes (updates push `content: null`; body saves do not touch the record row), and the pull engine fetched CRDT bodies only for `changedNoteIds`. Desktop peers see body edits because desktop probes every note's CRDT sequence each pass. Mobile now probes the 30-day recent window each pass (30 s floor), mirroring desktop; older bodies stay on-demand.
2. **An open app never re-pulled** — the only triggers were fg/online transitions and the BG task, so visibility of a desktop edit was set by the *next app switch* (uniform ~43 s in the drill). A 5 s foreground poll now runs while the app is active.

**20-trial drill results (desktop → staging → simulator, app foreground):** 0 misses; body median 32.2 s / p95 34.0 s; metadata median 28.1 s / p95 33.1 s (best trial 7.9 s). The floor is set by the DESKTOP push side: PushCoordinator drains the outbox and CRDT state on **30-second cycles** (`CrdtSnapshotScheduler` + push cycle), so <5 s end-to-end is not reachable for any edit type without making desktop pushes event-driven — flagging that as an owner decision (Phase 4 write-path work is the natural home).

Issue #1841, opened 2026-08-25
[bootstrap-sync] P3.1: Mobile adoption of the bootstrap flow


h4yfans:
Part of #1828 (Phase 3). Protocol was designed mobile-aware (settled Q6); this is the port.

- [ ] Bootstrap session + pack download + presigned attachment flow on mobile
- [ ] Mobile constraints: background transfer limits, disk pressure, cellular pacing (opt-in for large attachment prefetch on cellular)
- [ ] Reuses the desktop protocol unchanged — no mobile-only endpoints

Blocked by: #1840.

Issue #1840, opened 2026-08-25
[bootstrap-sync] P2.2: Bootstrap manifest + client pack download and stream-apply


h4yfans:
Part of #1828 (Phase 2).

- [ ] Bootstrap response (P1.2) includes the pack list (presigned) + tail cursor
- [ ] Client downloads packs in parallel at line speed (Range-resumable), newest cursor range first → recent notes appear first
- [ ] Stream-apply: decrypt + apply per item as pack bytes arrive, using the P0.3 page-transaction apply path; CRDT snapshots from packs seed Y.Docs without any snapshot GETs
- [ ] After packs: tail delta via existing item-granular `/sync/changes` + `/sync/pull`, then normal steady-state
- [ ] Fallback: no packs available (small vault / compaction lag) → current item-granular bootstrap, unchanged
- [ ] Compat: old clients ignore packs entirely and keep working

Blocked by: #1837, #1839 (bootstrap session + pack pipeline; apply path from #1831).

h4yfans:
Shipped in #1842 (merged 2026-08-26, `b9da51797`). It stayed open because that PR listed its issues as `Closes #1829, #1830, …`, and GitHub only binds the closing keyword to the first number in such a list. Re-checked against `main` at `f08e8a19b` on 2026-09-24.

- A fresh device lists `GET /sync/packs`, downloads presigned packs newest-first (3 in parallel), verifies them and stream-applies them. Snapshots seed Y.Docs without snapshot GETs, then the item-granular tail takes over. Any failure falls back to today's path with the cursor untouched.
- Known gaps: record packs (#1852) and pack fragmentation (#2380), which on a heavily edited vault can make the pack path cost more round trips than the item-granular path.

Issue #1839, opened 2026-08-25
[bootstrap-sync] P2.1: Pack compaction pipeline — queue-triggered, cursor-range immutable packs, derived cache


h4yfans:
Part of #1828 (Phase 2 — pack pipeline).

10k notes = 10k+ small encrypted blobs (record payloads + CRDT snapshots); even presigned, that's 10k RTTs. Kill the small-object problem by compaction — without touching E2E: a pack is a byte-concat of encrypted blobs + an index block; the server never decrypts anything.

- [ ] Cloudflare Queues consumer: after push batches land, compact new small blobs into cursor-range immutable pack files (~64–256MB target) in R2
- [ ] Pack contents: record payloads AND CRDT note-body snapshots (this is what kills the 10k-snapshot-GET floor)
- [ ] Pack index: D1 table mapping cursor range → pack key + per-item offsets
- [ ] Semantics: **derived cache** — individual blobs remain source of truth; no invalidation (deleted items simply remain as dead bytes in old packs; quota counts individual blobs only); tail is always item-granular
- [ ] Subrequest budget: consumer batch size chosen against the Worker subrequest ceiling (compaction does many R2 GETs per pack)
- [ ] Backfill job for existing vaults (paced)

Blocked by: #1836.

h4yfans:
Shipped in #1842 (merged 2026-08-26, `b9da51797`). It stayed open because that PR listed its issues as `Closes #1829, #1830, …`, and GitHub only binds the closing keyword to the first number in such a list. Re-checked against `main` at `f08e8a19b` on 2026-09-24.

- Queue-triggered compaction into immutable MPAK packs, the `pack_index` and `pack_watermarks` tables, a paced backfill, and a 6-hourly cron. Nudge retry was hardened in #2023. The production queue and consumer are live.
- Deviations: only `crdt_snapshot` is packed; record packs need signature metadata first (#1852). The target is 24 MB, not 64–256 MB, to fit the 128 MB isolate memory limit.
- Production data shows heavy fragmentation: 794 packs across 6 vaults, 3.7x superseded entries, and one vault with 527 packs for 299 live notes. Follow-up: #2380.

Issue #1838, opened 2026-08-25
[bootstrap-sync] P1.3: Move rate limiter off D1 — Workers rate-limit binding / Durable Object


h4yfans:
Part of #1828 (Phase 1).

The limiter is a D1-backed fixed-window counter: every rate-limited request pays a 2-statement `db.batch` against a single `rate_limits` row per bucket (`src/middleware/rate-limit.ts:45-56`) — write contention exactly when a fresh device hammers the API, plus 2 D1 ops of latency on every request.

- [ ] Migrate to the Workers rate-limiting binding (or a DO counter where per-device keying needs it)
- [ ] Preserve current bucket semantics (per-user, per-device for CRDT routes, Retry-After on 429)
- [ ] Bootstrap-session elevation (P1.2) hooks in here
- [ ] Drop the `rate_limits` table after migration window

h4yfans:
Shipped in #1842 (merged 2026-08-26, `b9da51797`). It stayed open because that PR listed its issues as `Closes #1829, #1830, …`, and GitHub only binds the closing keyword to the first number in such a list. Re-checked against `main` at `f08e8a19b` on 2026-09-24.

- The request limiter runs on the `RateLimiter` Durable Object (`durable-objects/rate-limiter.ts`, `middleware/rate-limit.ts`) with the same bucket semantics and Retry-After behaviour, plus the elevation hook for #1837.
- Task "drop the `rate_limits` table" does not apply: OTP (`services/otp.ts`) and the exception budget (`services/exception-budget.ts`) still use that table on low-volume paths, so it stays.

PR #1842, opened 2026-08-25
Bootstrap Sync epic (#1828): control-plane Worker, presigned R2 data plane, cursor-range packs

### LINKED ISSUE #1829: [bootstrap-sync] P0.1: Attachment download manager — queue, pacing, retry, resume, stream-to-disk, re-driver
Part of #1828 (Phase 0 — no protocol change).

Today the download side has none of the hardening the upload side has (`apps/desktop/src/main/sync/attachments.ts:508-634`, `ipc/sync-attachment-handlers.ts:444-509`):

- No download queue: per-emit spawn, unbounded concurrency → 429 storm + memory spike.
- No 429/Retry-After handling, no retry at all (`fetchManifest`/`downloadChunk` call `binaryFetch` bare) — one blip kills the whole file.
- No resume: any chunk failure discards all downloaded chunks; next attempt restarts at chunk 0.
- Files reassembled fully in RAM (`Buffer.concat` of all decrypted chunks) — parallel large files can OOM the main process.
- No re-driver: a failed download only retries when its note happens to be re-applied from a pull (`item-handlers/note-handler.ts:495,655`). Fresh devices end with silently missing attachments.

Build a proper download manager:

- [ ] Global download queue with bounded concurrency and pacing against the `blob_download` bucket (200/min today)
- [ ] 429 → honor Retry-After (global pause, like upload's `UploadQueue`), transient errors → per-item backoff via `withRetry`
- [ ] Chunk-level resume (keep landed chunks on disk, skip on retry)
- [ ] Stream chunks to a temp file, decrypt+verify per chunk, atomic rename — no full-file RAM buffering
- [ ] Persistent re-driver that walks `attachment-download-state` failures on reconnect/interval (independent of note re-apply)
- [ ] Priority ordering for the hybrid strategy (settled Q2): recently-used + small first; settings toggle for on-demand-only
- [ ] Resolve the eager/lazy contradiction: both `note-handler.ts:142-169` (eager emit on pull apply) and on-demand IPC paths exist — make the eager path feed the queue instead of firing directly

h4yfans:
## Summary

Epic #1828 — **Bootstrap Sync**. Makes a large vault interactive in seconds on a fresh device, with background download at line speed.

The Worker becomes a **control plane** (auth, manifest, bootstrap session, presigned URL issuance); R2 becomes the **data plane** (direct presigned GET/PUT). Note-body snapshots are compacted into cursor-range immutable **packs** (blind byte-concat of ciphertext + index block). Vault open is decoupled from sync. **The E2E encryption model is unchanged — the server still only ever handles ciphertext.**

64 commits, 203 files, +25819/−1411. Per-package branches are pushed for reference but must NOT be opened as separate PRs — conflict resolutions are baked into this branch.

### What shipped, per issue

- [x] #1838 — Rate limiter off D1 onto a `RateLimiter` Durable Object
- [x] #1833 — Server push batch rewrite (seeding lands as one cursor range)
- [x] #1834 — Sync-server quick wins (dead dedup + dead PUT removed, manifest pagination)
- [x] #1835 — Bootstrap telemetry (`time_to_interactive`, `time_to_full_text`, `bootstrap_bytes_per_sec`)
- [x] #1829 — Attachment download manager (queue, pacing, 429 handling, retry, resume, streams to disk)
- [x] #1830 — Vault open unblock + progressive open UX
- [x] #1831 — Apply-path performance (page transaction, worker decrypt, overlap prefetch)
- [x] #1832 — Background + incremental index build
- [x] #1836 — Presigned R2 GET/PUT (SigV4) + reclamation of chunks a session never registered
- [x] #1837 — Bootstrap session (elevated limit window, tail cursor)
- [x] #1839 — Pack compaction pipeline (queue-triggered, cursor-range immutable packs)
- [x] #1840 — Bootstrap manifest + client pack stream-apply

Mobile adoption (#1841) stays open — the protocol is mobile-aware by design, this implementation is desktop-first.

## Review findings fixed in this branch

A full-diff adversarial review (four independent reviewers with mutation testing, every finding then put to two skeptics) produced **19 candidates → 11 confirmed, 8 refuted**. All 11 are fixed. The two most serious:

- **The bootstrap router was mounted at `/sync`, not `/sync/bootstrap`** — the client's `POST /sync/bootstrap` 404'd and `openBootstrapSession` swallows failures by design, so **100% of #1837 was dead on arrival on every deploy, silently**. CI could not see it: `bootstrap.test.ts` builds its own app and mounts it at the intended path, so re-pointing the production mount to `/totally-wrong` still left 1208/1208 green.
- **A non-final pack hole aborted compaction permanently.** The hole branch never called `pack.skipEntry()`, so the next `writeEntry` threw before `advanceWatermark` — the queue dropped the message and the 6-hourly cron re-selected the same range forever, so that vault never packed again. The existing "tolerates holes" test only put the hole in the *last* slot.

Also fixed, found outside the review:

- **The packaged desktop build was broken.** `worker.ts` gained an import of `crdt-encrypt`, which pulled the crypto barrel (→ keytar, bip39, `require("electron")`) into the sync worker chunk. `pnpm build` failed at `check-worker-bundles` while every vitest suite stayed green. CI ran `electron-vite build` directly and skipped that check entirely — it is now its own CI step.
- **E2E had never been run for this epic and was broken.** #1838 made `RATE_LIMITER` a required, fail-closed binding, but the Miniflare harness never declared it, so every rate-limited route answered 500.
- **The bootstrap window and elevated session never closed on a throttled sweep.** The sweep throttle reads persisted `lastCrdtSweepAt` while fresh-device detection reads `lastCursor`, and two ordinary paths (`device-registration.ts`, `purgeKeyScopedSyncState`) delete the latter without the former — so an ordinary sign-in inside the sweep interval leaked the per-user session slot until its TTL and emitted no completion telemetry. Session release is now decoupled from the full-text claim and bounded by `BOOTSTRAP_DRAIN_BLOCKED_DWELL_MS`.
- **`bootstrapPullSucceeded` was vacuous.** `SyncEngine.pull()` swallows every error, so it was `true` on every run. `pull()` now reports whether it actually delivered.

## Deliberate scope decisions

**Record packs are not built.** `PACKED_KINDS` is `['crdt_snapshot']` only. A record pack entry carries just the encrypted payload; the Ed25519 signature, signer device id, clock and operation live in the `sync_items` row and reach a client only via `POST /sync/pull`, so no client can verify a packed record without skipping signature verification. Packs are immutable, so building them would mint R2 objects that are dead forever. The snapshot axis is where the per-note GET floor was. Tracked as #1852 (the entry meta is free-form JSON, so carrying the metadata later is additive, not a format break).

## Known open issues — read before merging

- **#1853 — attachment download never resumes when the socket is cut mid-body.** `binaryFetch` wraps only the header phase, so a stream truncated after the head flushed throws a raw `net::ERR_INCOMPLETE_CHUNKED_ENCODING` that `isResumableDownloadError` does not recognise; the verified prefix is deleted and the item dropped rather than re-queued. On a link that truncates mid-stream a large attachment never converges. **Not a regression** — before this epic there was no queue, retry, resume or re-driver at all and files were buffered whole in RAM — but the resume machinery is not reached for the commonest real interruption. A `test.fail` pin runs every suite run and turns red the moment it is fixed. Wrapping the two body reads was tried and is necessary but NOT sufficient; that attempt is recorded in the issue.
- **Seeding produces 429s.** Seeding a 200-note vault on one device produced 448 HTTP 429s, all on `POST /sync/crdt/snapshot` and `/sync/crdt/updates`. Elevation is pull-only by design, so the seeding half of the epic's scope is not accelerated. Not investigated as a regression against main.

## Accepted limitations (deliberate, documented)

- **#1831** — journal `writtenAt` is stamped per rewrite (self-heals via CRDT).
- **#1839** — a size mismatch, a vanished blob or a zero-length blob skips the hole rather than aborting. Dup-check keys on `min_cursor` alone, so a `created_at` tie-group split across windows leaves the second half permanently item-granular. `byteSize` counts the payload region only.
- **#1837** — a closed or stolen token can still elevate for ≤60 min from its own device (stateless elevation; TTL-bounded, pull-only, renew/close identity-bound).
- **#1834** — the paginated manifest walk consumes the `sync_manifest` bucket without client pacing (bounded, self-healing).
- `BootstrapOpenResponse.packs` is permanently empty — #1840 discovers packs through `GET /sync/packs` instead. Dead weight in the contract.

## Deploy prerequisites

1. **R2 presign (#1836)** — per-env secrets `R2_ACCESS_KEY_ID`, `R2_SECRET_ACCESS_KEY`, `R2_S3_ENDPOINT`, `R2_S3_BUCKET`. `R2_S3_BUCKET` **must match the STORAGE binding's bucket**; nothing validates this (a Worker cannot read a binding's bucket name), and a mismatch stays silent until `UPLOAD_INCOMPLETE` (400) at upload-complete. All absent ⇒ graceful proxied-only behavior.
2. **Bootstrap sessions (#1837)** — optional secret `BOOTSTRAP_SESSION_HMAC_KEY`. Absent ⇒ typed 501 + zero elevation = exactly today's behavior.
3. **Queues (#1839)** — `memry-pack-compaction-staging` / `-production`. The deploy workflows carry idempotent `wrangler queues create … || true` steps; create the queue before the first deploy, since a consumer bound to a missing queue fails the deploy.
4. **Migrations** — `0007_bootstrap_sessions.sql`, `0007_pack_index.sql`, `0008_upload_sessions_presigned_chunks.sql` apply on deploy. All additive; the two `0007`s are order-independent.
5. **Deploy order** — the sync-server must deploy **before** any desktop build emitting `sync_bootstrap`; one unknown event name 400s a whole telemetry batch.

Staging secrets and queues are configured. Production still needs its own set (separate R2 token, production bucket name).

## Compatibility — verified, not asserted

**Compat drill run and passed.** The `origin/main` desktop app (built at `d63ff48b0`) was run against **this branch's Worker** in Miniflare against the real migration ledger, by pointing the old worktree's harness at the epic's `apps/sync-server`. All five old-client sync specs passed — `shared-sync-bootstrap`, `account-sync` (×2), `manual-sync-smoke`, `dual-device-isolation`. The bundle was verified to be the new Worker (`listPacks`, `RateLimiter`, `bootstrap_sessions`, `pack_index`, `presignS3Url` all present; the old worktree's own server has zero occurrences of `listPacks`).

New client ↔ old server: `/sync/packs` 404, a missing/expired `url`, zero packs, or a schema-invalid response each fall back to today's item-granular bootstrap with the cursor untouched.

## Release note

Fresh-device setup is dramatically faster: your vault opens in seconds instead of minutes, and notes and attachments download in the background at full speed.

## Test plan

All run on `8f31fe564`:

- `pnpm lint` · `pnpm typecheck` · `pnpm check:architecture` · `pnpm check:contracts` · `pnpm ipc:generate && pnpm ipc:check` — pass
- `pnpm --filter @memry/sync-server test` — **1225 passed**
- `pnpm --filter @memry/desktop test:main` — **7461 passed**, 1 expected fail
- `pnpm --filter @memry/desktop test:renderer` — **8407 passed**
- `pnpm --filter @memry/desktop build` — passes, all five worker bundles clean
- `pnpm docs:impact --base origin/main --strict` · `pnpm docs:build` — pass
- New E2E: fresh-device bootstrap (200 notes, converges, zero 429s), seeding push (gap-free cursor range + old-client-shaped small batches), pack path (the Worker builds real packs in Miniflare and a fresh device bootstraps against them; a pack-less vault falls back), download-manager resume. One `test.fail` pin for #1853.

Closes #1829, #1830, #1831, #1832, #1833, #1834, #1835, #1836, #1837, #1838, #1839, #1840


h4yfans:
## Full-diff adversarial review — results

Four independent reviewers (fresh context, one per area, each in an isolated worktree with
reviewer-driven mutation testing) swept the whole `origin/main..sync/bootstrap-epic` diff.
Every candidate finding was then put to two skeptics with distinct lenses (reachability,
mechanical correctness) whose default position was to refute it.

**19 candidates → 11 confirmed, 8 refuted.**

### Confirmed

| # | Sev | Area | Finding |
|---|-----|------|---------|
| 1 | **CRITICAL** | server | Bootstrap router mounted at `/sync`, not `/sync/bootstrap` — every #1837 endpoint unreachable in production |
| 2 | HIGH | server | Pack-compaction declared-vs-actual size drift guard untested |
| 3 | **HIGH** | server | Hole/size-drift skip never calls `pack.skipEntry()` — any non-final hole aborts compaction and permanently stalls the vault |
| 4 | MEDIUM | crypto | Presigned chunk PUTs allow unbounded, unaccounted R2 writes that no sweep reclaims |
| 5 | MEDIUM | tests | Per-entry sha256 verification in `parsePack` untested |
| 6 | MEDIUM | tests | `pack-backfill` per-tick budget stop condition untested |
| 7 | LOW | client | Abort mid-apply commits a partial page, cursor still advances, run finalizes as success |
| 8 | LOW | tests | Bootstrap lazy-prune fixture never executed (missing `.run()`) — vacuous test |
| 9 | LOW | tests | `SyncWorkerBridge.decryptCrdtBatch` has zero direct coverage |
| 10 | LOW | tests | `bulk-apply` nested-BEGIN guard untested |
| 11 | LOW | tests | `listPacks` `hasMore` boundary untested — exact-multiple final page emits spurious `nextCursor` |

**#1 and #3 are the serious ones.**

#1: `index.ts:221` does `app.route('/sync', bootstrap)` while `routes/bootstrap.ts` registers
`'/'`, `'/renew'`, `'/close'` — so the shipped paths are `POST /sync`, `/sync/renew`,
`/sync/close`, and the client's `POST /sync/bootstrap` 404s. `openBootstrapSession` swallows the
failure by design, so **100% of #1837 (elevated pull ceilings, first manifest page, tailCursor,
chunk-hash page) was dead on arrival for every device on every deploy, silently.** Verified by
dumping the real `app.routes` table: zero routes contain the substring "bootstrap". CI could not
see it because `routes/bootstrap.test.ts` builds its own app and mounts it at the *intended*
path — re-pointing the production mount to `/totally-wrong` still left 1208/1208 green. It also
published an unintended authenticated `POST /sync`.

#3: the hole branch does `holes.push(plan.id); continue` without `pack.skipEntry(plan)`, so
`openPack`'s `nextSlot` never advances and the *next* `writeEntry` throws
`pack entries must be written in plan order`. The throw escapes before `advanceWatermark`, so the
queue retries 3× and drops, and the 6-hourly cron re-selects the identical range forever — the
watermark never moves and that vault never packs again. The existing "tolerates holes" test
passes only because it puts the hole in the *last* slot, where the ordering invariant is never
exercised. Note this also means the documented accepted-limitation ("size mismatch → hole-skip,
not abort") did not match what the code actually did.

### Refuted (8)

Worker CRDT-decrypt alignment vacuity · bootstrap renewal `expiresAt` NaN loop · presign-batch
ownership scoping unpinned · telemetry batch-poisoning blast radius · `MAX_SIZE_WALK_ENTRIES`
unbounded walk · bulk-apply re-entrancy guard untested · `MAX_PRESIGN_TTL_SECONDS` self-referential
clamp test · `DownloadPacer` production default unexercised.

### Separately: E2E was broken on this branch

`pnpm test:e2e` had never been run for this epic. #1838 made `RATE_LIMITER` a required binding
with an explicit no-fail-open path, but the Miniflare E2E harness
(`tests/sync-harness/src/simulated-server.ts`) never declared it — and the epic never touched
`tests/`. Every rate-limited route would 500 under E2E. The epic's own unit test
(*"should block the request when the RATE_LIMITER binding is absent"*) confirms the semantics.
Binding added. `R2_*`, `BOOTSTRAP_SESSION_HMAC_KEY` and `PACK_QUEUE` are all optional with
graceful degradation, so this was the only hard break.

### Baseline gates on `75d249cfe`

`lint` · `typecheck` · `check:architecture` · `check:contracts` · `ipc:check` — all pass.
sync-server 1208/1208 · desktop main 7356 · renderer 8407 · `docs:impact --strict` exit 0.
(One renderer failure on the first run was a load flake — `calendar-page.test.tsx` consent prompt,
unrelated to this epic; isolated rerun gave 8407 passed.)

Fixes for all 11 are in flight across four conflict-free branches, each with an independent
verifier that re-applies the original mutation to prove the new test actually pins the fix.

PR #1868, opened 2026-08-27
fix(scripts): exempt tooling configs from the mobile reachability walk

h4yfans:
check:architecture is red on current main:

```
- apps/mobile/vitest.config.ts -> node:url (node builtin reachable from apps/mobile)
```

`pnpm typecheck` runs the check, so every branch cut from main fails it. The mobile reachability rule (spec 001-mobile-app T003) guards the app's runtime import graph; a vitest config runs under node by definition and was only picked up because the walk skips test files but not tool configs. This skips `*.config.{ts,js,mjs,cjs}` in `walkMobileSources`, next to the existing test-file skip.

Verified: `node scripts/check-architecture-boundaries.js` fails on main, passes with this change; no other walk sites touched.

### DIFF
--- a/scripts/check-architecture-boundaries.js
+++ b/scripts/check-architecture-boundaries.js
@@ -96,6 +96,12 @@ function isTestFile(filePath) {
   return /\.(test|spec)\.(ts|tsx|js|jsx|mjs|cjs)$/.test(filePath)
 }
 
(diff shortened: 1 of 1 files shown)


h4yfans:
main independently landed the same tooling-config exemption; rebasing onto main leaves zero diff, so closing as redundant.

PR #1879, opened 2026-08-27
feat(tasks): link canvases from the Related picker

### LINKED ISSUE #1849: Task drawer "Related" cannot link a canvas
Part of #1843

## Reported

> Yesterday I setting a task, I didn't see canvas listed as one of the option in "Related" section in task side panel.

— Aurelie Kabore, 25 Aug 2026

Consistent with her 22 Aug report that "Recently opened" did not list canvases (fixed in v2026-08-23). Canvases keep being second-class in surfaces that enumerate vault items.

## Confirmed in code

The Related section is backed exclusively by notes. [task-detail-drawer.tsx](apps/desktop/src/renderer/src/components/tasks/task-detail-drawer.tsx):

```ts
void notesService.list({ sortBy: 'modified', sortOrder: 'desc', limit: 50 }).then((res) => {
  setAvailableNotes(res.notes.map((n) => ({ ... })))
})
```

and the link itself is stored as `task.linkedNoteIds`. A canvas is not a note — `FileType` is `'markdown' | 'pdf' | 'image' | 'audio' | 'video'` ([packages/shared/src/file-types.ts](packages/shared/src/file-types.ts)), with no canvas member. Canvas is a separate sync item type. So there is no path by which a canvas could appear in this picker.

The precedent for the right shape is already in the codebase — [packages/contracts/src/recents-api.ts](packages/contracts/src/recents-api.ts) models exactly this two-kind case:

```ts
export const RECENTLY_OPENED_ITEM_TYPES = ['note', 'canvas'] as const
```

That is the pattern the 22 Aug "Recently opened" fix established. Follow it.

## Also found while investigating — not reported

`limit: 50` with no server-side search: the picker fetches the 50 most recently modified notes **once**, and the search box then filters that array client-side. In a vault of any size, a note older than the 50 most recent is unfindable through this picker no matter what you type. Aurelie has not hit this, or has not reported it. Worth fixing in the same pass since the query is being touched anyway; split it out if it grows.

## Scope

- Extend the Related picker to list canvases alongside notes, with the canvas icon.
- Extend the task's link storage to hold a canvas reference.
- Clicking a linked canvas opens the canvas.

## Backward compatibility — required

`linkedNoteIds` is a persisted, **synced** field. Existing tasks carry note ids in it, written by older app versions, and older app versions will keep reading it.

- Do not repurpose `linkedNoteIds` to hold mixed ids — an older build would try to open a canvas id as a note.
- Prefer an additive field, or a typed-reference field alongside the existing one, with `linkedNoteIds` maintained for older clients.
- State the migration and compat plan on the PR before changing the shape. This crosses the sync protocol, the IPC contract, and the task DB schema.

## Done when

- A canvas can be linked to a task from the Related picker and opens on click.
- A task with a linked canvas, synced to a device running an older build, does not break or lose data there.
- A task with existing `linkedNoteIds` is unchanged after the migration.

h4yfans:
Closes #1849

## Behavior change

The task drawer's **Related** picker listed only notes, so a canvas could not be linked to a task at all. It now lists canvases alongside notes, each with its own icon, and clicking a linked canvas opens it in a canvas tab.

The same pass fixes the second defect the issue flags: the picker fetched the 50 most recently modified notes **once** and filtered that array in memory, so a note outside that window was unfindable no matter what you typed. Note results now come from the existing FTS index (`search:query`), so the whole vault is reachable. The empty-query case still lists recent notes, so opening the picker looks the same as before.

## Storage design

`linkedNoteIds` is not a column — it is the `task_notes` junction table. The canvas half mirrors it as a new `task_canvases` junction, with `linkedCanvasIds` added to the sync payload beside `linkedNoteIds`. `linkedNoteIds` is untouched and no canvas id ever enters it.

Two alternatives were rejected on backward-compatibility grounds, not taste:

| Option | Old-client round-trip | Verdict |
|---|---|---|
| `linkedCanvasIds` as a JSON column on `tasks` | **Wiped.** The plain-apply branch writes scalar columns unguarded (`description: data.description ?? null`), so an absent key clears the column | Rejected |
| Polymorphic `task_links(task_id, item_type, item_id)` | Breaks the guard — an older client writing `linkedNoteIds` would union-merge against a table it can only see half of. Also forces migrating existing `task_notes` rows | Rejected |
| **`task_canvases` junction mirroring `task_notes`** | **Preserved.** Junction writes are guarded, so an absent key means "unchanged" | **Chosen** |

At the UI layer the two disjoint fields are unioned into a discriminated reference (`{ kind: 'note' | 'canvas'; id }`) with `kind:id` React keys. A note and a canvas can share an id, so a bare `string[]` would have been ambiguous.

`linkedCanvasIds` is deliberately **not** added to `TASK_SYNCABLE_FIELDS`. Junction relations carry no field clocks; they are union-merged, exactly like `tags` and `linkedNoteIds`.

## Migration

`apps/desktop/src/main/database/drizzle-data/0055_task_canvases.sql`, hand-written (the Drizzle generator has been off since 0020; `db:generate` was not run). Journal entry 55 hand-appended, no meta snapshot.

Additive only: one `CREATE TABLE IF NOT EXISTS`, no `ALTER`, no backfill, no `DELETE`. Existing installs upgrade by gaining an empty table; every `tasks` row and every `task_notes` row is left exactly as it is.

It mirrors `task_notes` column for column **including the absence of a foreign key to `canvases`**. A FK there would make sync apply order load-bearing: a task arriving before the canvas it links to would fail the whole upsert. The `ON DELETE CASCADE` FK to `tasks` stays so deleting a task reclaims its links.

## Older-client analysis

The issue asked whether a round-trip through an old client drops the new field. Traced end to end:

1. **Old client receives a task carrying `linkedCanvasIds`.** `TaskSyncPayloadSchema` is a plain non-strict `z.object`, so Zod **strips** the unknown key. The task applies normally — it does not reject. This matters: a schema failure returns `'skipped'`, which advances the cursor permanently. The schema is deliberately left non-strict.
2. **Old client pushes the task back.** It builds its payload from its own DB and has no canvas rows, so the key is **absent entirely** — not empty.
3. **New client applies that payload.** Every canvas write sits behind `if (data.linkedCanvasIds)`, so an absent key preserves the local rows. Canvas links survive the round-trip.

An explicit `linkedCanvasIds: []` still clears, which is how a real unlink travels. Absent means "no information"; empty means "none". That distinction is what makes the round-trip safe, so it is covered by tests in both the merge and plain-apply branches.

Rollback is inert: an older build never reads `task_canvases`, and its migrator's max journal `when` is lower, so it applies nothing and never drops the table.

**One inherited property worth naming:** junction merges are unions, so a concurrent unlink loses to a link. Canvas links behave exactly like note links here. This is pre-existing for notes and was matched rather than diverged from; changing it needs per-link tombstones and belongs in its own issue.

## Verification

- **Reviewer-side mutation testing.** Removing the guard on the plain-apply branch fails exactly `preserves local canvas links when an older client omits linkedCanvasIds on APPLY` and nothing else. Removing the merge-branch guard and union fails exactly the two MERGE tests. The guards are load-bearing and the tests prove it.
- Main-process: 4 files, 54 tests passing, covering the migration upgrade path (`leaves an existing task and its note links untouched when upgrading`, idempotent re-execution, `cascades on task delete and has no foreign key to canvases`), the three compat tests above, union-merge, and both push paths.
- Renderer: 35 tests passing, including `lists notes and canvases together in the related picker` and `links a canvas from the picker and opens the canvas when its row is clicked`.
- `pnpm lint` 0 · desktop `typecheck:web` / `typecheck:node` / `typecheck:test` 0 · `pnpm ipc:generate && pnpm ipc:check` 0 with no regenerated diff · `git diff --check` clean · `docs:impact --strict` and `docs:build` 0.
- Nothing was added to the `tsconfig.test.*` exclude backlog.

## Notes for the reviewer

- `pnpm ipc:generate` needs Node 24 (`--experimental-transform-types` was removed in Node 26); the native modules are built for the Node running the tests.
- Repo-wide `pnpm typecheck` currently fails on `apps/mobile/vitest.config.ts -> node:url`, an architecture-boundary violation present on `origin/main` and untouched by this branch. `main` is independently red on Desktop CI and an E2E shard as of the #1863 merge.
- Out of scope, deliberately: the kanban card badge and `task-linked-note-indicator` still count only note links, so a task linked solely to a canvas shows no badge. Cosmetic, and widening those is a separate change.


h4yfans:
---

## Follow-ups after the first CI run

**A regression CI caught that my targeted test run had missed.** `enrichTask` now reads the `task_canvases` junction, and `tasks-handlers.test.ts` partially mocks `@main/database/queries/tasks`. The mock did not list `getTaskCanvasIds`/`setTaskCanvases`, so all 25 handler tests failed on the missing export. Fixed in `b0f356092`; the GET assertion now passes a real canvas id so the enrichment is covered rather than defaulted to `[]`. `Unit & integration tests` passes on the following run.

**Test coverage for the new hook** (`a33fc2fe7`), prompted by the Codecov comment. The hook's two sources settle independently so one failing cannot blank the other — that invariant was stated in a comment and nothing tested it.

Worth flagging how that test was validated, because the first two versions of it passed against deliberately broken code. `mockRejectedValue` settles before the sibling source's `.then` chain does, so a handler that wrongly cleared its sibling's results had its damage overwritten before the assertion ran. The failure cases now reject on a timer, which reproduces the realistic ordering; mutating the hook to clear the sibling fails exactly that test and nothing else.

**React Doctor's `transition-all` warning is a false positive on "new".** `transition-all` appears twice in this file on `origin/main` and twice here — unchanged. It is flagged because the union refactor moved the line. Changing it would not trace to this issue, so it is left alone.

**CI status.** Everything this PR owns is green. The one red check, `Lint, typecheck, test, and boundary check`, is a pre-existing `main` breakage this branch does not touch — `apps/mobile/vitest.config.ts -> node:url`, present identically on `origin/main`. Fixing it revealed a second one behind it (a stale `editor-web` build artifact). Both are fixed in #1882; this PR goes green once that lands.

PR #1882, opened 2026-08-28
fix(ci): stop the mobile boundary gate from auditing build-time configs

h4yfans:
`main` has been red since the #1863 merge:

```
architecture boundary check failed:
- apps/mobile/vitest.config.ts -> node:url (node builtin reachable from apps/mobile)
```

Every open PR inherits it, so nothing can go fully green until this lands.

## Why the gate is asking the wrong question here

The mobile reachability rule keeps node builtins and electron out of anything the React Native bundle can reach. `apps/mobile/vitest.config.ts` imports `node:url` to build its `@ -> src` alias. It is the test runner's own config: it runs under Node by definition and never ships in the RN bundle, so it is tooling, not reachable app code.

The walker already encodes this distinction — it skips the `scripts` directory and skips test files. Build-time configs are the same category and were simply missed.

## Why not rewrite the config instead

Every portable way to resolve that alias needs a node builtin: `fileURLToPath` needs `node:url`, `path.resolve` needs `node:path`. The one import-free shortcut, `new URL('./src', import.meta.url).pathname`, yields `/C:/...` on Windows and would break the Windows jobs. Dropping the alias is worse — the config comments explain it exists so tests can reach modules by the same `@/*` mapping the app uses.

## Change

`walkMobileSources` now skips files matching `*.config.{js,ts,cjs,mjs,cts,mts}`, alongside the existing test-file skip.

Narrow by design: it excludes only files whose basename ends in `.config.*`, so real RN sources are still audited. The rule still catches a node builtin in any actual mobile source or in any workspace package reachable from one.

## Verification

`pnpm check:architecture` fails on `main` with the error above and passes with this change.


h4yfans:
---

## Second breakage, uncovered by the first fix

With the boundary step passing, the job advanced to a step that had never been reached and failed there too:

```
editor-web asset is STALE.
  expected eb1b6ccc87652bf5
  found    bd7e457e72a3d587
```

`apps/mobile/src/editor/generated/editor-web-asset.ts` is a committed build artifact whose stored contract hash no longer matches its sources, so a source change landed without its regenerated bundle.

Regenerated with the command the gate itself names, `pnpm --filter @memry/mobile editor:build`. The output hash is `eb1b6ccc87652bf5` — exactly what CI computed as expected — so the build is reproducible and this is the artifact that should have been committed alongside the source change. The diff is the hash line plus the packed blob.

Both `pnpm check:architecture` and `pnpm --filter @memry/mobile editor:check` pass on this branch.

h4yfans:
Rebased onto current `main` and regenerated. The asset now hashes to `c042ae553c3b2886`, which is what CI computes as expected.

**Why the first regeneration did not stick, and why this will keep happening.** The freshness hash covers `packages/contracts/src/webview-bridge.ts`, `apps/mobile/editor-web`, and `packages/editor-schema/src`. That last one is the problem: three editor fixes landed on `main` between my base and now — #1866, #1875, #1876 — and all three touched `packages/editor-schema/src` without regenerating the mobile asset. Each one moved the expected hash.

So the asset is not stale because of one missed regeneration. Any change to the shared editor schema invalidates it, and the desktop-side editor work that touches that package routinely ships without rebuilding the mobile bundle. This will go red again on the next `editor-schema` change.

Worth considering separately (not in this PR): either regenerate the asset in CI rather than committing it, or add the rebuild to the same gate that catches it, so the fix does not depend on the author of an unrelated desktop editor fix knowing about the mobile bundle.

Issue #1892, opened 2026-08-30
Blank-line gaps adjacent to a callout or a quote collapse on the renderer's write-back


h4yfans:
Part of #1843. Found while fixing #1877 (the same defect on the toggle splitter), by widening the mixed-document fuzz family to join its parts on a gap instead of a plain paragraph break.

## What happens

An extra blank line next to a `> [!info]` callout — or next to a structured `>` quote run — collapses on the **renderer** pipeline. The main/CRDT pipeline keeps it.

```
> [!info]
> Body


After
```

comes back with a single blank line before `After`. Write-back byte-compares, so the file is rewritten on the next open and the user's spacing is gone.

## Why

Exactly #1877's shape, one splitter over. `splitMarkdownByBlockquoteRuns` (`apps/desktop/src/renderer/src/components/note/content-area/callout-block.tsx`) runs BEFORE `splitMarkdownPreservingBlanks`, and its `flushMarkdown` does `mdLines.join('\n').trim()` — the gap lines at a callout's or a quote's edge are trimmed out of the markdown segments before the blank-line scanner ever sees them.

The main pipeline is unaffected because it reads callout and quote runs line by line inside `parseContentWithColorMarkers`, i.e. AFTER `splitMarkdownPreservingBlanks` has already lifted the gaps out.

#1877's fix is the shape of the answer: `splitMarkdownByToggles` now emits a `gap` segment for the blank lines at each of its seams, and its two callers turn one into that many empty paragraphs.

## Repro

Fuzz family `mixed documents` in the #1848 conformance suite, marked `pending: { renderer: <this issue> }` since #1877's PR widened its join to `\n\n\n`. The static case `extra blank line next to a callout survives` in the same suite pins the minimal shape.

## Done when

The pending flags come off the fuzz family and the static case, and both pipelines round-trip gaps adjacent to callouts and quotes byte-identically.

PR #1901, opened 2026-08-30
fix(editor): keep the bytes around a toggle and the ones a declined toggle owns

### LINKED ISSUE #1877: Blank-line gaps adjacent to a toggle collapse on write-back
Part of #1843, found by the mixed-document fuzz family of the #1848 conformance suite.

## What happens

An empty paragraph next to a toggle disappears on the first save. `Alpha\n\n\nBeta` round-trips byte-identically, but the same extra blank line collapses whenever one side is a `<details data-memry-toggle>` region — before it, after it, and between two toggles:

```
<details data-memry-toggle>...</details>\n\n\nAfter   →  <details ...></details>\n\nAfter
Before\n\n\n<details ...>                          →  Before\n\n<details ...>
<details ...>\n\n\n<details ...>                   →  <details ...>\n\n<details ...>
```

Write-back byte-compares, so the file is rewritten on next open and the user's deliberate spacing is gone. Identical behavior on both pipelines (renderer `parseMaskedMarkdown` and main `blocknote-converter.ts`).

## Why

`splitMarkdownByToggles` (`packages/editor-schema/src/blocks/markdown.ts`) runs BEFORE the blank-line-preserving scanner, and its `flushMarkdown` does `buffer.join('\n').trim()` — the gap lines around a toggle region are trimmed out of the markdown segments before `splitMarkdownPreservingBlanks` ever sees them. On the way back out, segments rejoin with a fixed `'\n\n'`.

## Repro

Conformance case `extra blank line next to a toggle survives` in the #1848 suite (marked `it.fails` pending this issue), plus the three-line probe above through `markdownToYFragment` → `yDocToMarkdown`.

## Done when

The pending flag comes off the conformance case and both pipelines round-trip gaps adjacent to toggles byte-identically.

### LINKED ISSUE #1883: Unterminated toggle drops its open and summary lines on write-back
Part of #1843, pinned by the #1848 conformance suite. #1876 persisted fold state but did not reach this shape; #1847's investigation section named it.

## What happens

A `<details data-memry-toggle>` whose `</details>` is missing — a half-typed toggle, a truncated sync payload — round-trips to just its body text on both pipelines:

```
<details data-memry-toggle>
<summary>Unterminated</summary>      →      Body

Body
```

`splitMarkdownByToggles` correctly declines the region ("an unterminated toggle stays markdown"), but the leftover raw-HTML lines then reach BlockNote's markdown parser, which has no block for them and drops them — the toggle marker and the summary text are gone from the file for good on the next write-back.

## Repro

Conformance case `unterminated toggle stays literal markdown` in the #1848 suite (marked `it.fails` pending this issue): `<details data-memry-toggle>\n<summary>Unterminated</summary>\n\nBody` → `Body` through both `markdownToYFragment` → `yDocToMarkdown` and the renderer pipeline.

## Done when

The pending flag comes off and the unterminated shape round-trips byte-identically — the decline has to actually preserve the declined bytes.

h4yfans:
## Summary

Closes #1877
Closes #1883

Two write-back defects on one code path, `splitMarkdownByToggles` in `packages/editor-schema/src/blocks/markdown.ts`. Both destroy real bytes in real vault files: write-back byte-compares, so any drift rewrites the note on its next open.

**The failing inputs**

```
#1877  Before\n\n\n<toggle>\n\n\nAfter   ->  Before\n\n<toggle>\n\nAfter
       <toggleA>\n\n\n<toggleB>          ->  <toggleA>\n\n<toggleB>

#1883  <details data-memry-toggle>
       <summary>Unterminated</summary>   ->  Body
       (blank)
       Body
```

**Root cause, #1877.** The toggle split has to run before the blank-line and marker-line scanners, because those read one line at a time and would shred a toggle body at its own paragraph gaps. Running first is also what caused the bug: nothing downstream ever sees the lines at a toggle's edges, so those gaps were this function's to carry and it threw them away. `flushMarkdown` did `buffer.join('\n').trim()`, and the colors-marker scan popped the trailing blanks on its way past — two separate places, which is why fixing only the `trim()` left the gap-before-a-toggle case still broken. The splitter now emits a `gap` segment per seam, in the same `extraLines` currency `splitMarkdownPreservingBlanks` already uses inside a segment, and both callers turn one into that many empty paragraphs using the code they already had for a mid-segment gap. Serialization was never involved and is untouched.

A gap before the *first* segment is still dropped. Measured rather than assumed: assembly writes a gap as `\n\n` plus its extras whether or not a segment precedes it, so with nothing in front those are two more blank lines, and three leading blank lines come back as four, four as five, growing on every save. Trimming them is main's behavior and it stays.

**Root cause, #1883 — and the part worth a reviewer's attention.** The issue says `splitMarkdownByToggles` "correctly declines the region", and the function's own docstring says an unterminated toggle "stays markdown". Both were measured and both are wrong about what declining *produces*. Declining is correct; it just did not preserve anything. The declined lines are raw HTML, and BlockNote's markdown parser has no block for raw HTML, so it dropped them and the open and summary lines were gone from the file for good.

Every `<details>` / `<summary>` / `</details>` line the splitter declines is now escaped with `\<` on its way into the parser. CommonMark reads it as text, and remark writes the backslash back out as nothing, so the author's bytes come back unchanged. Every `<` on the line is escaped, not just the leading one — a `<summary>x</summary>` whose closing tag stays raw comes back as `<summary>x`.

**Chosen behavior for an unterminated toggle.** It stays literal markdown, byte-for-byte, and is not adopted as a toggle. The alternative — treating a missing `</details>` as running to end of document — swallows the rest of the note into a block the author never closed and adds a `</details>` the author never wrote, so it changes the bytes either way. The note shows the markup as literal text, which is honest about the file being malformed, and closing the block by hand turns it into a real toggle on the next open.

**This rule is narrower than #1883 and fixes more.** A hand-written Obsidian `<details>` was being destroyed by the same parser, despite the file's standing promise that a block without our attribute is left as its author wrote it. That promise is now true. `raw-html-roundtrip.probe.test.ts` is the repo's instrument for this and it records the move: four probes leave "a block-level tag is dropped, its inner text kept" for "what survives".

**Both serializer copies.** Both share the fixed splitter, so the defect and the fix are identical on each; the only per-copy change is the four-line gap branch in each `parseMaskedMarkdown`. Both halves of the conformance corpus assert the same canonical bytes.

**Compatibility.** No new node type, no schema change, no new props. A note written here still opens on an older install. The mobile editor asset is rebuilt because it packs `packages/editor-schema/src`.

**Found one construct over, filed not widened.** The renderer's `splitMarkdownByBlockquoteRuns` has the identical `.trim()` at a callout's or a quote's edge — #1892. The main pipeline is unaffected there because it reads those runs after the blank-line scanner, not before. The mixed-document fuzz family now joins its parts on a gap instead of a paragraph break, which is what caught it; it is green on main and pending on the renderer half.

## Release note

Notes keep the blank lines you put around a collapsible toggle, instead of closing the gap on the next save. A `<details>` section written by hand, or one whose closing tag is missing, is now left in the file exactly as written rather than being dropped.

## Test plan

Nine corpus cases were added first, failing, in `5f15b39fb`, then the fix turned them green. Red run with only the three source files reverted to `origin/main`, corpus and fuzz and golden fixture all failing:

```
main pipeline       Tests  11 failed | 58 passed | 1 expected fail (70)
renderer pipeline   Tests   9 failed | 50 passed | 5 expected fail (64)

'unterminated toggle stays literal markdown'
'extra blank line next to a toggle survives'
'extra blank line between two toggles survives'
'two extra blank lines before a toggle survive'
'extra blank line before a colored toggle survives'
'unterminated toggle with no summary keeps its open line'
'unterminated expanded toggle keeps its open and summary lines'
'foreign details block stays the bytes its author wrote'
'orphan closing details tag stays literal markdown'
'mixed documents' stay byte-identical across 48 seeded cases
roundtrip-toggles.md body survives the converter round-trip byte-identical
```

Green, rebased on `a4756d520`:

```
pnpm exec vitest --project main  roundtrip + golden + raw-html + converter
  Test Files  4 passed (4)        Tests  422 passed | 1 expected fail (423)
pnpm exec vitest --project renderer  roundtrip-conformance
  Test Files  1 passed (1)        Tests   59 passed | 5 expected fail (64)
pnpm --filter @memry/editor-schema test
  Test Files  6 passed (6)        Tests  189 passed (189)

BUILD_BEFORE_TEST=1 pnpm --filter @memry/desktop test:e2e toggle-writeback-byte-stability
  2 passed (9.2m)
    blank lines on both sides of a toggle survive a real save (#1877)
    an unterminated toggle keeps its open and summary lines through a save (#1883)

pnpm lint                 0 errors, 109 warnings (all pre-existing, none in changed files)
pnpm typecheck            19 successful, 19 total
pnpm check:architecture   architecture boundary check passed
git diff --check          clean
pnpm docs:impact --strict covered
pnpm docs:build           build complete
pnpm --filter @memry/mobile editor:check   asset is current (2f5d69cd52810f8b)
```

The E2E asserts the vault `.md` with `fs`, not through `getNoteFileBodyById` or the write-back debug state's `lastMarkdown` — both run through `normalizeBodyText`, which collapses `\n{3,}` to `\n\n`, precisely the byte difference #1877 is about, so either would have reported the bug fixed while the file still lost the gap.

Full-suite runs show pre-existing reds unrelated to this change (module-mock failures and 30s timeouts in importers, crypto, linking-service, calendar, notes-tree). All 12 such files pass in isolation with this branch applied: main 9/9, renderer 3/3. `notes-tree-folder-rename-duplicate.test.tsx` is a known load flake that fails identically on clean `origin/main`.


h4yfans:
CodeQL alert 174 fixed in `c152c1117` — real, not a nitpick.

`escapeDetailsMarkup` escaped only `<`, so a backslash already in front of one paired with the escape being added. `<summary>C:\<path></summary>` went into the parser as `...C:\\<path>...`, CommonMark read the `\\` as a single literal backslash, and `<path>` was raw again for the parser that drops raw HTML. The line lost its tag on write-back, which is precisely the loss the helper exists to prevent. A summary ending in a backslash was worse: the escape landed on `</summary>`'s own bracket and the whole closing tag went. Both were also unstable, degrading further on each subsequent save.

Fix is backslashes first, then brackets:

```ts
if (!DETAILS_MARKUP_LINE_REGEX.test(line)) return line
return line.replace(/\\/g, '\\\\').replace(/</g, '\\<')
```

**Measured, not assumed**, because doubling every backslash could have over-escaped lines whose backslashes are nowhere near a bracket. It does not: CommonMark reads `\\` back as one backslash and remark writes a literal backslash bare, so the doubling on the way in is exactly undone on the way out. Round-trips through the real pipeline, before → after:

```
backslash before <       DRIFT UNSTABLE  ->  IDENTITY stable
backslash ending line    DRIFT UNSTABLE  ->  IDENTITY stable
doubled backslash        DRIFT stable    ->  IDENTITY stable
backslash elsewhere      IDENTITY stable ->  IDENTITY stable   (guard, unchanged)
windows path C:\Users\me IDENTITY stable ->  IDENTITY stable   (guard, unchanged)
```

Three corpus cases on both pipelines. The two `data-memry-toggle` ones are red against the old escape; the foreign-block one passes on both sides on purpose, as the guard that fails if the escaping ever over-reaches.

```
red   (old escape)   main  Tests  2 failed | 70 passed | 1 expected fail (73)
                     rend  Tests  2 failed | 60 passed | 5 expected fail (67)
green                main  Tests  152 passed | 1 expected fail (153)   (roundtrip + golden + raw-html)
                     rend  Tests   62 passed | 5 expected fail (67)
pnpm --filter @memry/editor-schema test   189 passed (189)
pnpm typecheck            19 successful, 19 total
pnpm lint                 0 errors, 109 warnings (pre-existing, none in changed files)
pnpm check:architecture   passed
git diff --check          clean
pnpm docs:impact --strict covered
editor:check              asset is current (a5c2918d12ed7785)
```

Note the pre-existing behaviours this does *not* change, all app-wide remark canonicalizations outside this corpus's contract and identical on `origin/main`: in an ordinary paragraph `a\\b` collapses to `a\b`, a trailing backslash is dropped, and `a \* b` becomes `a * b`.

PR #1918, opened 2026-08-30
fix(sync): stop rewriting markdown files Memry did not author

### LINKED ISSUE #1909: [Bug]: Memry rewrites any foreign markdown file it opens into its own dialect — 12 of 14 ordinary inputs mutate
Part of #1906

## Report

> and that file is corrupted now in Obsidian. which happened to be a kanban board, so you definitely cant just point it to an obsidian vault, or you run the risk of destroyng your files. very poor

`v2026-08-25` · Win32, through the in-app form. Anonymous — the fix is the only answer this person gets.

## What is actually happening

**This is not about Kanban.** Memry rewrites *any* markdown file it did not author into its own dialect. Kanban is simply the one case where the user could see it, because a machine reads that file rather than a human.

Measured on 2026-08-31 against `markdownToYFragment` → `yDocToMarkdown` — the exact pair the app runs, seeding the doc on note open and re-serializing the whole doc on the first write-back after it. 14 ordinary markdown inputs, **12 came back different**:

| Input | Output | Loss |
| --- | --- | --- |
| `Line one··\nLine two` (two-space hard break) | `Line one\n\nLine two` | **semantic** — a `<br>` becomes a paragraph break |
| `[docs][d]` + `[d]: https://example.com` | `[docs](https://example.com)` | **semantic** — the reference definition is deleted |
| ```` ``` ```` untagged fence | ```` ```javascript ```` | **semantic** — a shell or JSON block now claims to be JavaScript |
| `* One` | `- One` | cosmetic |
| `+ One` | `- One` | cosmetic |
| `_em_` / `__strong__` | `*em*` / `**strong**` | cosmetic |
| `Title\n=====` (setext) | `# Title` | cosmetic |
| `---` thematic break | `***` | cosmetic |
| `Text:\n- Item` | `Text:\n\n- Item` | cosmetic |
| 4-space nested list indent | 2-space | cosmetic |
| `# Title ` (trailing space) | `# Title` | cosmetic |
| `**Bold**\n- Item` | `**Bold**\n\n- Item` | cosmetic |

Only two inputs survived byte-identical: a language-tagged fence, and an ordinal list starting at 3.

A second pass over the output is stable — this is a one-time mutation per file, not runaway drift.

### Why Kanban is the visible case

An Obsidian Kanban board carries a `%% kanban:settings %%` block holding fenced JSON. Round-tripping a real one injects a blank line after `%% kanban:settings`, tags the fence `javascript`, and injects another blank line before the closing `%%`. For a block parsed by a plugin rather than read by a person, "cosmetic" does not apply.

### Why it matters beyond one file

The user pointed Memry at an existing vault. Every file Memry opens comes back rewritten. On a large vault that is a working tree where nothing is as its author left it — and the three semantic rows above are real losses buried inside that noise.

## Reproduce

Drop this at `apps/desktop/src/main/sync/tmp-probe.test.ts`, run it, then delete it. Node 24 — Node 26 breaks the tooling.

```ts
import { describe, it } from 'vitest'
import * as Y from 'yjs'
import { CRDT_FRAGMENT_NAME } from '@memry/contracts/ipc-crdt'
import { markdownToYFragment, yDocToMarkdown } from './blocknote-converter'

async function onePass(md: string): Promise<string> {
  const doc = new Y.Doc()
  await markdownToYFragment(md, doc.getXmlFragment(CRDT_FRAGMENT_NAME))
  return (await yDocToMarkdown(doc)) as string
}

const CASES: Array<[string, string]> = [
  ['hard line break', 'Line one  \nLine two'],
  ['reference link', 'See [the docs][d].\n\n[d]: https://example.com'],
  ['untagged fence', 'Run this:\n\n```\nnpm install\n```'],
  ['asterisk bullets', '* One\n* Two'],
  ['underscore emphasis', 'This is _em_ and __strong__.'],
  ['setext heading', 'My Title\n========\n\nBody.']
]

describe('foreign markdown probe', () => {
  it('what survives', async () => {
    for (const [name, md] of CASES) {
      const out = await onePass(md)
      console.log(`${name} :: ${out === md ? 'UNCHANGED' : 'MUTATED'}`)
      if (out !== md) console.log('  in : ' + JSON.stringify(md) + '\n  out: ' + JSON.stringify(out))
    }
  })
})
```

```bash
pnpm --filter @memry/desktop exec vitest run --config config/vitest.config.ts --project main src/main/sync/tmp-probe.test.ts
```

## Proven, and not proven

**Proven.** The byte-level mutations in the table. The pipeline pair that produces them. Stability on a second pass.

**Not proven, and worth an hour before writing code:**

1. **When the write actually fires.** The measurements show what happens to a body that goes through the pipeline. They do not show what causes a file the user never opened to go through it. `53c1672e0` ("the write-back stops overwriting bytes it has never read") added a `contentHash` guard and its own message names the gap it left: *"A note whose hash was never measured — a tier-0 sidebar row, listed from `stat` alone — has nothing to compare and writes as it always did."* Establish whether that is the path here, or whether the user simply opened the file.
2. **Whether the Kanban plugin actually breaks.** The bytes change; nobody has fed the mutated file back to the plugin. Do that before repeating the word "corrupted" to a customer.
3. **Frontmatter is a red herring.** A first probe showed `---` becoming `***` — that was an artifact of feeding raw file bytes to the converter. `crdt-writeback.ts:524` parses and re-serializes frontmatter separately via `parseNote` / `serializeParsedNote`. Do not chase it.

## The decision this issue exists to make

Three options, smallest first. Pick one deliberately and record why.

**A — Never write a file whose hash was never measured.** Closes the tier-0 gap from `53c1672e0`. Small, safe, and stops the bleeding for files the user never opens. Does nothing for a file they do open.

**B — Fix only the three semantic rows.** Hard line breaks, reference definitions, untagged fences. Small, and it is the subset that actually loses information. Leaves the vault-wide rewrite noise in place.

**C — Byte-identical round-trip for anything the pipeline did not author.** The real fix and the expensive one. This is the promise `markdown.ts` already makes for a bare `<details>` (`b0a020ea0`) generalised to the whole file.

A and B together are cheap and cover the user's actual loss. C is the honest answer and probably its own epic. **A + B is the recommendation**; open C as a follow-up rather than growing this issue into it.

## Acceptance criteria

Scoped to A + B. Restate these if a different option is chosen.

- A file Memry has never read is never overwritten by the write-back. No exception for tier-0 rows.
- A two-space hard line break survives a round-trip as a hard line break.
- A reference-style link and its definition both survive; a definition referenced more than once is not inlined and deleted.
- An untagged code fence comes back untagged. No language is invented.
- Conformance cases join the existing suite in the shape `ca14df318` and `b0a020ea0` use: pin the bytes, fail on current main, then fix.
- E2E reads the file, the shared doc and the UI in one poll, so a failure names the layer rather than leaving it to a screenshot.

## Backward compatibility

Live beta on real vaults. Tightening the write-back must not stop legitimate saves — if the hash guard is widened, prove with a test that a note the user is actively editing still writes. A file already mutated by a previous version must not be mutated further.

## Related

Same vein, already fixed: #1877 (blank-line gaps collapse), #1883 (raw `<details>` dropped), #1892 (gaps around callouts and quotes), `b0a020ea0`, `ca14df318`.

Siblings in #1906: #1907 (imported Obsidian task cannot be edited) and #1908 (import Obsidian Tasks plugin lines). #1908 is blocked by whatever this issue decides about writing into foreign files.

## Verify

```bash
pnpm test
pnpm test:e2e
pnpm typecheck
```

h4yfans:
## Summary

Memry rewrote any markdown file it did not author into its own dialect. Twelve of fourteen ordinary inputs came back changed, three of them losing meaning rather than style. This lands options A + B from the issue: the write-back stops touching files it has never read, and the three information-losing rows are fixed.

**A. Never write over bytes this app has not read.** `53c1672e0` added the `contentHash` compare and named the hole it left: a row listed from `stat` alone has no hash, so the guard had nothing to compare and wrote anyway. That hole is how a vault was rewritten with nobody opening a file. `FullSyncRunner.sweepAllCrdtNotes` queues a pull for every markdown note in the vault after a reconnect and in the `finally` of every full sync, pack bootstrap does the same on a fresh device, and the inbound update applies with `ORIGIN_NETWORK`, which schedules a write-back that re-serializes the whole document and writes the whole file. A row that exists with no hash is now refused, with no exception for tier-0.

Refusing alone would have cost a real save, because nothing else fills that column in: `indexVault` skips a path that already has a row, so a null hash can be permanent. So `CrdtProvider.seedFromMarkdown` now records the hash of the bytes it built the document from. Opening a note is what makes the guard let go. A row that is not in the index at all is untouched: `cached` then came from canonical metadata, which means the item handler applied this note and wrote this file moments ago and only the projection is late.

**B. The three rows that lose information.**

- **Hard line break.** The document never lost it: the editor holds a soft break as one newline inside a text node and a hard break as two, and remark writes one backslash break for the first and two for the second. `normalizeSerializedMarkdown` collapsed every backslash break to a plain newline, turning that pair into a blank line. A run of exactly two is now written as a two-space hard break; runs of one and of three or more keep the old collapse.
- **Reference-style links.** The definition is a block BlockNote has no node for, so it was dropped and the destination inlined at every use site. Definitions are now lifted out before the editor sees the markdown and ride beside the document in two Y.Arrays, the same shape CriticMarkup marks use, with each use site's exact spelling recorded. On the way out each usage rewrites the first inline link matching its text and destination, so an inline link that merely shares a destination is left alone, and a definition several links share is restored at all of them. Definitions come back at the end of the file: CommonMark resolves one wherever it sits, so this costs the position of a mid-file definition and loses nothing.
- **Untagged code fences.** A bare fence parses to a `codeBlock` whose language is the schema default, `javascript`. The fences are now counted off the source and matched against the parsed code blocks in document order, and only the invented language is cleared. A count mismatch abandons the whole pass rather than guessing.

## The two open questions the issue asked to settle first

**When the write actually fires.** Both paths are real, and neither needed a user edit. Seeding a document does not schedule a write-back: `blocksToYFragment` transacts with a `null` origin and `crdt-provider.ts:1218` gates on `ORIGIN_NETWORK || isIpcOrigin`, which rejects it, and on the `doOpen` path the update handler is attached after the seed anyway. A note the user never opened is reached by the vault-wide sweep above. A note the user opens with zero typing is rewritten on the first open by the four `replaceBlocks` promoters at `use-editor-sync.ts:388-393`, but only when the body holds a `[[...]]`, a non-list `[ ]`, `((date:...))` or `((mention:...))` that main seeded as plain text. `contentHash` is populated by `indexVault` for every file present when the vault opens, so the null-hash window is narrow: a file that appears while the app runs and whose tier-1 backfill is cancelled by a vault close within 500ms keeps a null hash forever, because `indexer.ts:139-143` skips a path that already has a row.

**Whether an Obsidian Kanban board genuinely breaks.** Yes, and exactly one of the three changes causes it. `extractSettingsFooter` (`src/parsers/parseMarkdown.ts:39-63`) is a reverse character scanner, not an AST walk, and line 60 does `JSON.parse(md.slice(i + 1, settingsEnd).trim())` where `i + 1` is the character straight after the opening fence's third backtick. The info string lands inside the slice, so `javascript\n{"kanban-plugin":"basic"}` throws. `StateManager.getParsedBoard` catches it and returns an empty board carrying the error, and `Kanban.tsx:188-202` replaces every lane and card with a bare `Error:` heading and a stack trace. `saveToDisk` is gated while errors exist, so the file itself is not clobbered further, but the board is unusable until someone strips the tag by hand. The two blank lines the round trip also injects are harmless: phase one of that scan tolerates `` ` ``, `%`, `\n` and `\r`. The shipped 2.0.51 `main.js` carries the byte-identical minified scanner, and the function has never been rewritten since it landed in 2021. Those two blank lines are therefore pinned as canonical rather than chased.

## Migration and backward compatibility

Live beta on real vaults, so nothing here resets or migrates anything.

- **No schema change.** `note_cache.content_hash` is an existing nullable column. The seed only ever fills a hole in it and never overwrites a hash the indexer measured, so no migration is needed and an older build reading the same row is unaffected.
- **No editor schema change and no new node type.** The two reference arrays are additive Y.Arrays. A build that predates this ignores them and behaves exactly as it does today; a note written here still opens on an older install.
- **A file an older build already mutated is not mutated again.** The index measured the mutated bytes, so the hashes agree and the write is a no-op against them. Pinned by a test.
- **Legitimate saves still write.** Widening the guard is paired with the seed recording, and a test asserts the edit of a note whose hash was filled in by the seed still reaches disk.
- **The renderer's own serializer gets the hard-break half** through the shared `normalizeSerializedMarkdown`. The reference-link and fence halves are main-pipeline only, which is the pipeline the write-back runs; porting them to the renderer parse path is a follow-up.

## Release note

Files written in another app are no longer rewritten. memrynote will not touch a note it has never opened, and opening one now keeps its hard line breaks, its reference-style links and their definitions, and code fences that were given no language. Obsidian Kanban boards keep working.

## Test plan

Node 24 throughout.

- `pnpm --filter @memry/desktop test:main` — 7720 passed, 1 expected fail, 6 skipped, 565 files. Exit 0.
- `pnpm --filter @memry/shared exec vitest run` — 268 passed, 16 files. Exit 0.
- `pnpm --filter @memry/desktop test:renderer` — 8543 passed, 2 expected fail, 7 skipped. One timeout in `notes-tree-folder-rename-duplicate.test.tsx` under the parallel load of three concurrent suites; that file passes 14/14 on its own and is unrelated to markdown. Exit 1 on the batch, exit 0 in isolation.
- `pnpm typecheck` — exit 0.
- `pnpm lint` — exit 0, no new warnings in any changed file.
- `git diff --check` — exit 0.
- `pnpm docs:impact --base 6aea0a0c0 --strict` — covered.
- `pnpm docs:build` — exit 0.

Unit tests added:

- `apps/desktop/src/main/sync/foreign-markdown-roundtrip.test.ts`, 21 cases in the shape `blocknote-converter.roundtrip.test.ts` uses. 18 of them fail on `origin/main`; all 21 pass here.
- `packages/shared/src/link-references.test.ts`, 19 cases over the strip and restore halves.
- `apps/desktop/src/main/sync/crdt-writeback.test.ts`, 3 new cases: a file this app has not read is never written; a note whose hash the seed filled in still saves; a file a previous version already mutated is left alone.

E2E added: `apps/desktop/tests/e2e/foreign-markdown-roundtrip.e2e.ts`, 3 cases, one per information-losing row. Each seeds a foreign file, opens it, edits, waits for write-back, closes it, and reads the file bytes, the shared Y.Doc and the UI in one poll so a failure names the layer. **Not run locally** — E2E runs serially under the coordinator because it needs an Electron native rebuild. The expected values are reasoned from the converter diff and the unit sibling rather than observed; the three to watch on the first real run are the `doc` field in case 1, the anchor list in case 2, and the code block's `language` in case 3.

## Deliberate scope notes

- Option C, a byte-identical round trip for anything the pipeline did not author, is not attempted here and should be its own issue. The cosmetic rows are unchanged: `*` and `+` bullets still become `-`, `_em_` becomes `*em*`, setext headings become ATX, and `~~~` becomes ```` ``` ````.
- The shared corpus at `packages/editor-schema/src/conformance.ts` was outside this change's write scope, so the new cases live in a sibling main-pipeline file instead. Porting them there so the renderer half asserts them too is a follow-up.
- `feedExternalEditToCrdt` diffs text into the document without re-parsing, so the reference arrays go stale after an external edit to a note holding reference links. The definitions are still restored, just from the last parse. Worth a follow-up.

Closes #1909


h4yfans:
Docs gate waived with `MEMRY_DOCS_IMPACT_SKIP=1` on the CodeQL commit: it is an internal ReDoS fix in `packages/shared/src/link-references.ts` with identical observable behaviour, so there is nothing for `apps/docs/src` to describe.

PR #1919, opened 2026-08-30
fix(tasks): never dress a checkbox with no task row as a task

### LINKED ISSUE #1907: [Bug]: A task imported from an Obsidian file cannot be edited, re-projected or deleted
Part of #1906

## Report

> please let us change details of a task. i have tasks that have clearly come from an obsidian file. now i cant edit the task, change the project, delete the project.

Reported anonymously through the in-app feedback form on `v2026-08-25` · Win32 · not signed in. No reply address, so this cannot be answered directly — the fix is the answer.

## Why it happens

Memry's task identity is a suffix, not a checkbox. `packages/shared/src/task-block.ts:6`:

> A task is stored in markdown as a checkbox with a trailing `{task:<id>}` suffix, e.g. `- [ ] Buy milk {task:abc}`.

`scanTaskCheckboxStates` bails on the first line of `packages/shared/src/task-block.ts:65`:

```ts
if (!markdown.includes(TASK_BLOCK_SUFFIX_OPEN)) return states
```

A line written by Obsidian — `- [ ] Buy milk 📅 2026-09-01 ⏫` — carries no `{task:` suffix, so no `tasks` row is ever created for it and `reconcileTaskCheckboxesFromMarkdown` has nothing to reconcile. The markdown importer does not help either: `packages/importers/src/markdown/` is `map-files`, `parse-frontmatter` and `extract-asset-refs` only, with no checkbox handling anywhere in it.

The result is a checkbox that renders like a task and behaves like inert text. Every task affordance — edit, assign a project, remove the project — needs a row that does not exist.

## What to work out first

The report says "i have tasks", so something is surfacing them. Establish which of these is true before writing a fix:

1. The lines render as task blocks in the note body but never reach the Tasks page.
2. They reach the Tasks page through some other path and the edit affordances are separately broken.
3. They hold a `{task:<id>}` whose id resolves to no row (a vault copied between installs), and the unresolved-id branch at `apps/desktop/src/main/tasks/reconcile-markdown-tasks.ts:60` silently skips them.

Reproduce with a real Obsidian file before choosing. Branch 3 is cheap to test: hand-write a `{task:nonexistent}` suffix into a note and open it.

## Acceptance criteria

- A checkbox line in a note that Memry has no `tasks` row for is never presented with task affordances it cannot honour. Either it becomes a real task or it renders as a plain checklist item.
- Whichever way it resolves, clicking a task-looking row never leaves the user with a control that does nothing.
- A regression test covers a note body containing checkbox lines with no `{task:}` suffix, and one with an unresolvable id.

## Out of scope

Parsing Obsidian Tasks metadata — that is the sibling issue. This one is about never showing a dead control.

## Verify

```bash
pnpm typecheck
```

h4yfans:
## Summary

A checkbox line imported from Obsidian showed up as a task whose every control did nothing. The reporter put it plainly: *"i have tasks that have clearly come from an obsidian file. now i cant edit the task, change the project, delete the project."*

**Root cause.** An unsuffixed checkbox line does not stay a checkbox. `ContentArea.tsx` feeds `analyzeTaskIntents`' `standaloneCandidate` into a 600 ms debounce and then `convertCheckboxToTask`, which creates a real `tasks` row — which is why the lines reached the Tasks page at all. But the conversion rewrites the block to a `taskBlock` with `taskId: ''` **before** the row exists, then writes the id back afterwards. Three exits skipped that write-back and left the block at `taskId: ''` permanently: no project to create into, `tasksService.create` returning `{success:false}`, or a throw. The block could not be re-offered either, because the intent analyzer only picks `checkListItem`s as candidates. `convertCheckboxToSubtask` had the same three exits.

`TaskBlockRenderer` renders any unresolved `taskBlock` through `placeholderTask`, whose `id` is `''`. So the row carried a full set of live-looking controls whose handlers all early-return on the empty id: toggle (`if (!taskIdArg) return`), status and priority (`if (!taskId) return`), project (`if (!taskId) return`), and the open-in-task-panel arrow. Nothing was logged and no error was shown, because none of those check the envelope.

The same placeholder row shows for a `{task:<id>}` copied from another install, until the async lookup settles into the existing "Task deleted" notice. Verified with a probe: first paint is `<div data-testid="task-row" data-task-id="">`, and only after the lookup resolves does the notice replace it. `useTaskBlockData` exposes `isLoading`, but the renderer destructures it as `_isLoading` and ignores it, so it never distinguished "still loading" from "resolved to nothing".

**Fix.**

- `ContentArea.tsx` gets a `restoreCheckbox` helper. Every failure exit in both conversions puts the block back to `checkListItem` with its original content and tick. A checkbox with no row stays a checkbox — the acceptance criterion directly. The block stays in `dismissedBlocksRef` so a failing conversion does not retry in a hot loop; reopening the note retries with a fresh set.
- `task-row.tsx` takes `interactive`, default `true`. When false the status and priority pickers get `disabled` (both already supported it), the project badge renders through the existing static branch instead of `InteractiveProjectBadge`, and the row leaves the tab order with `aria-busy`.
- `task-block-renderer.tsx` passes `interactive={!!task}` and drops the navigate arrow when there is no task. The row keeps its shape so notes do not flicker on open, but offers no affordance it cannot honour.

The main-process skip at `reconcile-markdown-tasks.ts:60` is correct and unchanged: an id that names no row has genuinely nothing to reconcile. It is now covered by a test that says so.

Diagnostic write-up with the full evidence trail is on the issue: https://github.com/memrynote/memry/issues/1907#issuecomment-5472000658

## Release note

An imported checklist line that memrynote has no task for no longer appears as a task you cannot edit. It stays a plain checkbox until it becomes a real task, and a `{task:…}` naming a task from another install says so instead of pretending.

## Test plan

Backward compatibility: no schema, IPC contract, settings or file-format change. Vaults that already hold a `taskBlock` stuck at `taskId: ''` from an older build now render it inert rather than falsely live, and `createTaskForDraftBlock` can still heal it. Note files are not rewritten by any of this. `interactive` defaults to `true`, so every existing `TaskRow` call site is unchanged.

Unit tests, written first and run red against `origin/main` before the fix:

- `apps/desktop/src/renderer/src/components/note/content-area/task-block/unresolved-task-affordances.test.tsx` — 5 cases. Three were red (`expected 'true' to be 'false'`, `expected <button> to be null`); the two characterising the deleted-task notice and the resolved-task row were green from the start and guard against over-fixing.
- `apps/desktop/src/renderer/src/components/note/content-area/ContentArea.test.tsx` — 2 new cases, both red (`expected 'taskBlock' to be 'checkListItem'`), one per conversion failure path.
- `apps/desktop/src/renderer/src/components/tasks/task-row-interactive.test.tsx` — 5 cases on the real `TaskRow`, covering the new prop in both states.
- `apps/desktop/src/main/tasks/reconcile-markdown-tasks.test.ts` — 2 new cases pinning both imported shapes. Characterisation, green from the start.

E2E: `apps/desktop/tests/e2e/imported-task-affordances.e2e.ts`, two tests — one for checkbox lines with no `{task:}` suffix, one for an unresolvable `{task:<id>}`. **Not run locally**: the coordinator runs E2E serially because it needs an Electron native rebuild. It respects the suite's rules — `BUILD_BEFORE_TEST=1`, `dismissFirstRunOnboarding` in the seed, fields asserted one at a time rather than through `toMatchObject`, no synthetic load.

Commands run, Node v24.16.0:

```
pnpm --filter @memry/desktop test:main       # 7698 passed | 1 expected fail | 6 skipped (567 files)
pnpm --filter @memry/desktop test:renderer   # 8556 passed | 2 expected fail | 7 skipped (703 files)
pnpm lint                                    # exit 0
pnpm typecheck                               # exit 0
pnpm --filter @memry/desktop typecheck:test  # exit 0
git diff --check                             # clean
pnpm docs:impact --base 6aea0a0c0 --strict   # covered
pnpm docs:build                              # build complete
```

Closes #1907


h4yfans:
Independent review of this branch. Verified against the code, not the PR text. Two commits pushed on top.

**Confirmed and fixed.**

1. `convertCheckboxToTask` had a fourth failure exit the fix missed. `tasksService.listProjects()` was awaited outside the `try`, and an IPC invoke rejects when the main handler throws (vault closed mid-flight, DB error). That path left the block as a `taskBlock` at `taskId: ''` — the exact dead row this PR removes, still alive on one exit. The `try` now opens before the lookup; every failure path runs `restoreCheckbox`. Red-first test added in `ContentArea.test.tsx` (`reverts a checkbox to a checklist item when the project lookup rejects`), confirmed red before the fix and green after.

2. The E2E caveat was accurate: both specs pass on `origin/main`. A clean vault always has a project (`ensureDefaultTaskProject` runs on every vault open and `deleteProject` refuses the inbox), so the first spec never hits a failure exit, and the unresolvable-id end state was already correct on main. Made it a genuine repro: a `deleteAllTaskProjectsForE2E` test hook (gated behind `NODE_ENV === 'test'` like every other hook) deletes the project rows out from under the open app, so the next conversion trips the `tasks.project_id` foreign key deterministically. The new third spec asserts the lines stay plain checkboxes — red on main by construction, and it is also the only place `restoreCheckbox` runs against real BlockNote rather than the mocked editor.

**Verified and dismissed.**

- `interactive` defaulting to `true`: the real `TaskRow` has exactly two consumers, `task-block-renderer.tsx` (passes it explicitly) and `tasks-widget.tsx` (maps over tasks fetched from the DB, so a confirmed row backs every render). The other "TaskRow" files are separate components. No call site renders over an unconfirmed task.
- Recoverability after `restoreCheckbox`: the restored line is a native checklist item — tick, edit, everything works. It stays in `dismissedBlocksRef` until the note remounts, which is what the docs promise; retyping the line also mints a fresh block id and retries immediately. Not a new dead end.
- The `!defaultProject` exit is unreachable in normal use (inbox guaranteed at vault open, undeletable), so no error toast is warranted there; the reachable failure is the create itself, which the docs' "stays a plain checklist item, tries again next open" line covers. Leaving it silent is right.
- Claim that the unresolvable-id case was already correct on main: true (`get` → null → `isDeleted` → notice with a working remove button). The first-paint transient is genuinely closed by `interactive={!!task}`, which subsumes the ignored `isLoading` — no live control ever renders over the placeholder's empty id.
- `placeholderTask`: local to the renderer, still only feeds the non-interactive row. Nothing depends on it being live.

**Noted, deliberately left.** `createTaskForDraftBlock` keeps the same three leaky exits, but a draft block has no checkbox to restore, the renderer now renders a stuck draft inert instead of falsely live, and editing the title re-triggers the heal. Separate concern if it ever bites.

**Docs-impact skip on the push:** the flagged files are `test-hooks.ts` (test-only seam) and `ContentArea.tsx`, whose change extends behavior `capturing.md` in this branch already documents. No new user-facing surface.

Gates on the final head, Node v24.16.0: `test:main` 7698 passed | 1 expected fail (a `metadata.test.ts` teardown flake appeared once under full parallel load, clean in isolation and on rerun), `test:renderer` 8557 passed | 2 expected fail, `lint`, `typecheck`, `typecheck:test`, `git diff --check` all clean.

PR #1922, opened 2026-08-31
feat(tasks): import Obsidian Tasks plugin lines as first-class Memry tasks

### LINKED ISSUE #1908: [Feature]: Import Obsidian Tasks plugin lines as first-class Memry tasks
Part of #1906

## Why

A user pointed Memry at a vault whose tasks are managed by the **Obsidian Tasks plugin** (`obsidian-tasks-plugin`, currently v8.4.0, the standard choice for tasks in Obsidian). Memry imported the notes and dropped every task on the floor, because it only recognises its own `{task:<id>}` suffix.

Anyone arriving from Obsidian with tasks arrives in exactly this state. This is the import path that makes an Obsidian migration real rather than partial.

## The source format

The plugin ships two serializers. A reference clone sits at `~/workspace/obsidian-tasks`; read `src/TaskSerializer/DefaultTaskSerializer.ts` and `src/TaskSerializer/DataviewTaskSerializer.ts` rather than trusting this summary.

**Default (emoji) format** — `DEFAULT_SYMBOLS` in `src/TaskSerializer/DefaultTaskSerializer.ts`:

| Field | Symbol | Value |
| --- | --- | --- |
| priority | `🔺` `⏫` `🔼` `🔽` `⏬` | highest → lowest, none absent |
| start date | `🛫` | `YYYY-MM-DD` |
| created date | `➕` | `YYYY-MM-DD` |
| scheduled date | `⏳` (also `⌛`) | `YYYY-MM-DD` |
| due date | `📅` (also `📆` `🗓`) | `YYYY-MM-DD` |
| done date | `✅` | `YYYY-MM-DD` |
| cancelled date | `❌` | `YYYY-MM-DD` |
| recurrence | `🔁` | `[a-zA-Z0-9, !]+`, e.g. `every week` |
| on completion | `🏁` | `[a-zA-Z]+` |
| id | `🆔` | `[a-zA-Z0-9-_]+` |
| depends on | `⛔` | comma-separated ids |

Each symbol may carry an optional Variant Selector 16 (`️`) and the regexes are end-anchored and stripped from the tail one at a time — see `fieldRegex`. A trailing `^blockid` block link may follow.

**Dataview format** — `DATAVIEW_SYMBOLS` in `src/TaskSerializer/DataviewTaskSerializer.ts`: the same fields as inline fields, `[due:: 2026-09-01]`, `[priority:: high]`, `[repeat:: every week]`, `[completion:: ...]`, `[cancelled:: ...]`, `[start:: ...]`, `[scheduled:: ...]`, `[created:: ...]`, `[id:: ...]`, `[dependsOn:: ...]`. Note the field name differences: `completion` not `done`, `repeat` not `recurrence`.

Both formats hang off a normal list checkbox, and the plugin supports custom status characters beyond `[ ]` and `[x]` (`src/Statuses/StatusRegistry.ts`) — `[/]` in progress, `[-]` cancelled and whatever the user has configured.

Tags are plain `#tag` in the description. A "global filter" tag may be configured, in which case only lines carrying it are tasks.

## What to build

Read an Obsidian markdown file, recognise task lines in either format, and produce a Memry task the user can actually edit — a `tasks` row plus the `{task:<id>}` suffix written back into the line, so the two stay reconciled by the existing `reconcileTaskCheckboxesFromMarkdown` path.

Field mapping is the substance of this work. Decide and write down, per field, whether Memry has a home for it, and what happens when it does not:

- due / scheduled / start / done / cancelled dates → Memry's task date fields
- priority → Memry's priority scale
- recurrence → Memry's recurrence, or dropped with the original text preserved
- `🆔` / `⛔` dependencies → Memry has no task dependencies today; decide explicitly
- custom status characters → Memry's statuses
- `#tags` in the description → Memry tags, case-preserving

**Never silently drop a field.** A dropped field must either stay as literal text in the title or be recorded somewhere the user can find. Losing someone's due dates on import is worse than not importing.

## Backward compatibility

Existing installs must be unaffected: a vault with no Obsidian task lines behaves exactly as it does today, and a re-import of an already-imported file must not duplicate tasks. Rewriting a line to add `{task:<id>}` is a write into a user's vault file — it must go through whatever write path respects `contentHash`, and it must be idempotent.

## Acceptance criteria

- Both serializer formats parse, including Variant Selector 16, the alternate due/scheduled symbols, and custom status characters.
- A parsed task appears in the Tasks page and is fully editable: title, project, dates, priority, delete.
- Round-trip: import, edit in Memry, and the Obsidian file still reads correctly in Obsidian.
- Re-running the import over the same file creates no duplicates.
- Fixture tests over real plugin output, not hand-written approximations. Borrow cases from `~/workspace/obsidian-tasks/tests/TaskSerializer`.

## Verify

```bash
pnpm test
pnpm typecheck
pnpm check:architecture
```

h4yfans:
Anyone arriving from an Obsidian vault whose tasks are managed by the Obsidian Tasks plugin got their notes imported and every task dropped, because Memry only recognised its own `{task:<id>}` suffix. This makes those lines first-class Memry tasks.

Stacked on #1918.

## Where the import happens, and why there

Memry has no "import an Obsidian vault" wizard. You point Memry at the folder, and a markdown checkbox becomes a `tasks` row the first time you open that note, through the converter that already exists in `ContentArea.tsx`. This change extends that converter rather than adding a second path.

That placement falls out of #1918. Opening a note is the only moment Memry has genuinely read the file's bytes and recorded a `contentHash`. The `{task:<id>}` write-back already travels the CRDT to `crdt-writeback.ts`, which refuses to overwrite a file whose hash it never measured. So this change adds no write path and inherits that guard unchanged. Nothing under `apps/desktop/src/main/sync/` is touched.

## The constraint that shapes every decision

Memry's task identity is a trailing suffix: `serializeTaskBlock` writes `- [x] <title> {task:<id>}`, and `parseTaskBlockSuffix` only reads a suffix at the very end of the line. Every Obsidian Tasks field regex, and the plugin's block-link regex, is end-anchored (`$`) and stripped from the tail one at a time. Appending Memry's suffix to a line that still carries plugin metadata therefore un-anchors all of it, and Obsidian silently stops seeing its own fields.

So Memry lifts the fields it can carry off the line and into real task fields, and for the constructs it cannot move it leaves the line alone.

## Field mapping

| Obsidian Tasks field | Symbol | Memry home | Why |
| --- | --- | --- | --- |
| description | | task title | tags stay inline, as the plugin keeps them |
| priority | `🔺` `⏫` `🔼` `🔽` `⏬` | `priority` 4, 3, 2, 1, 1 | five plugin levels onto four Memry ones; `⏬` lowest collapses onto Low and the original line records which it was |
| due date | `📅` `📆` `🗓` | `dueDate` | direct |
| start date | `🛫` | `startDate` | direct |
| scheduled date | `⏳` `⌛` | `startDate` when there is no `🛫`, else preserved | Memry has one non-due date field |
| done date | `✅` | `completedAt`, and the box is ticked | built at local midnight on the written date; a date the calendar does not have is left on the description |
| cancelled date | `❌` | preserved | no column |
| created date | `➕` | preserved | `TaskCreateInput` hardcodes now |
| recurrence | `🔁` | `repeatConfig` + `repeatFrom`, else preserved | see the supported subset below |
| on completion | `🏁` | preserved | no equivalent |
| tags | `#tag` | `task_tags`, original casing kept | matches the case-preserving rule `setTaskTags` enforces; clamped to the 50 characters and 20 tags `TaskCreateSchema` accepts |
| id | `🆔` | **line left alone** | other files' `⛔` point at this id; deleting it breaks a graph spanning files Memry has never read |
| depends on | `⛔` | **line left alone** | Memry has no task dependencies, and the value is unrecoverable once stripped |
| block link | `^blockid` | **line left alone** | other notes link to it, and it must stay last on the line |
| custom status | `[/]` `[-]` … | **not converted** | see below |

**Every imported task keeps its original line, verbatim, on `description`.** Not just the ones carrying a field with no Memry home. Importing rewrites the line in the user's vault, and mapping `📅 2026-09-01` into a due date still loses their symbol choice, field order and spacing. A record of only the fields we happened to find useful is not a record. No labels, no prose, no new translated strings, and the line copies straight back into Obsidian to undo the import by hand.

**Recurrence** maps for `every day`, `every N days`, and the same for weeks, months and years, plus `every weekday` and day lists like `every Monday, Wednesday`. A trailing `when done` sets `repeatFrom: 'completion'`. Any richer rrule keeps its text and imports without a repeat, because a half-understood rule that fires on the wrong days is worse than no rule.

**Custom status characters are declined structurally, not by new code.** Standard markdown only treats `[ ]`, `[x]` and `[X]` as task list items. Verified against the installed `remark-gfm@4.0.1`: `- [/] in progress` parses with `checked: null` and keeps `[/] in progress` as literal text, so it never becomes a checkbox block and the converter never sees it. `serializeTaskBlock` can only write `[ ]` or `[x]`, so converting one would flatten the user's status character. The parser recognises all of them and the fixture tests cover them; the import declines them.

I also ran the app's own `markdownToYFragment` / `yDocToMarkdown` pair over the declined shapes rather than trusting that reasoning. `- [/] in progress`, `- [-] cancelled`, `- [ ] Do first 🆔 dcf64c` and `- [ ] Linked ^abc123` all come back byte-identical, so "left alone" is measured, not asserted.

**The global filter is declined, visibly.** The plugin can be configured so only lines carrying a tag (commonly `#task`) are tasks. Memry does not read the plugin's settings and treats every checkbox as a candidate, which is what it already did before this change. The filter tag survives as an ordinary Memry tag. This is documented as a known limitation.

## Backward compatibility and idempotence

Live beta, real vaults. No schema change, no contract change, no migration, no file-format change. `packages/shared/package.json` gains one `exports` subpath and nothing else.

A vault with no Obsidian task lines behaves exactly as it did: `buildObsidianTaskImport` returns null when the parser finds no plugin field, and the converter falls through to the existing `parseQuickAdd` path unchanged.

### Memry and the plugin both want the end of the line

An adversarial review of the design found a bug I would otherwise have shipped, and `486ace0df` fixes it. Memry's suffix has to be last for `parseTaskBlockSuffix` to find it, but the plugin appends its own field after whatever is already there. So completing an imported task back in Obsidian writes `- [x] Buy milk {task:abc} ✅ 2026-09-05`, and a strictly end-anchored read returned null: Memry lost the id it had written itself, the block regressed to a bare checkbox, `scanTaskCheckboxStates` missed the completion, and the next open minted a duplicate task with a garbled title. Ticking a box in Obsidian was enough to trigger it.

A suffix is now still ours when everything behind it is plugin syntax, and only then. `Buy milk {task:abc} and eggs` is unchanged, so ordinary trailing prose still means this is not a task line. Memry absorbs the plugin's edit, marks the task complete, and drops the trailing `✅` on the next save because the completion now lives on the task.

**Idempotence is structural and predates this change.** Once a line carries `{task:<id>}`, `normalizeTaskBlocks` upgrades it to a `taskBlock` and `hasTaskSuffix` in `analyzeTaskIntents` stops it ever being a conversion candidate again. Re-opening the note, or re-running the import over the same file, creates no second task. The E2E spec opens the note twice and asserts the count is unchanged, because that criterion is the one a single-import test would silently fake.

The three declined constructs leave their line byte-identical on disk. They are refused inside the converters themselves, before any block mutation, so the refusal holds however the conversion was asked for.

## Review pass on the wiring

A review of the first version found four data-loss paths, all in the wiring rather than in either tested unit. Each is fixed and pinned by a test that failed before the fix.

**The context menu bypassed the refusal.** `handleEditorContextMenu` calls `convertCheckboxToTask` directly and never consults `analyzeTaskIntents`, so the analyzer's refusal never ran. A line carrying `🆔` plus a due date converted, and the id another note's `⛔` names was gone from disk and absent from the database. The refusal now lives in both converters, where the rewrite happens, and the user gets a toast naming what to remove.

**A done date was reverted on the next save.** The converter completed the task but left `checked` false, so the block serialized as `- [ ] Buy milk {task:id}`. `reconcileTaskCheckboxesFromMarkdown` runs on every note save and every watcher event, saw `checked === false` against a non-null `completedAt`, and un-completed the task. The imported done date did not survive one save. The box is now ticked to match the completion, which is also what the plugin's own `✅` means.

**A nested line lost its fields.** `convertCheckboxToSubtask` passed the raw text as the title, so a nested `- [ ] Book flight 📅 2026-01-01` serialized as `- [ ] Book flight 📅 2026-01-01 {task:id}`: the suffix un-anchored the plugin's due-date regex, and Memry had not read the date either. The nested path now lifts the same fields as the top-level one.

**A half-converted block could be orphaned on `taskId: ''`.** The block becomes a `taskBlock` before the create call resolves, and the bare `catch` only cleared the dismissal. Two inputs reached it. A done date the calendar does not have, `✅ 2026-02-30`, parses as a date and `z.string().datetime()` rejects it, after `create` has already succeeded. A tag longer than the 50 characters `TaskCreateSchema` takes rejects the whole create, because the Obsidian tag grammar is looser than the contract. Both are now refused at the import boundary, a rejected completion no longer costs the block its task id, and anything else that fails puts the block back as the checkbox the file has.

`completedAt` was also a UTC-midnight instant built from a local calendar date, so `✅ 2026-01-07` rendered as the 6th anywhere west of UTC. It is built at local midnight now, for the same reason `formatDateKey` is local.

## Rejected alternatives

**Regenerate the plugin tail at write-back time**, keyed by task id, so the line reads `- [ ] Buy milk {task:abc} 📅 2026-09-01` with the suffix mid-line and the fields still end-anchored. This is the only option that keeps Obsidian fully functional. Rejected here because it needs `blocksToMarkdownPreserving` in `apps/desktop/src/main/sync/`, which this PR must not touch, and because it turns a checkbox-only reconcile into a two-writer field-level merge. Worth its own issue.

**A `trailer` prop on the `taskBlock` config** so `serializeTaskBlock` can re-emit the tail. Needs `packages/editor-schema`, out of scope, and it is a block-schema change with its own contract test.

**Use the plugin's `🆔` as Memry's identity.** Hijacks a field the user may already be using for dependencies.

## Verified

`pnpm --filter @memry/shared exec vitest run` 391 passed. `pnpm --filter @memry/desktop test:renderer` 702 files, 8593 passed. `pnpm typecheck`, `pnpm lint`, `pnpm check:architecture`, `pnpm --filter @memry/desktop typecheck:test`, `git diff --check` all exit 0. `pnpm docs:build` clean, `pnpm docs:impact --strict` covered.

`pnpm --filter @memry/desktop test:main` is 7724 passed with one failure, `fts-corruption.test.ts`, which passes on its own (`Tests 8 passed`). It is the known parallel-load flake and touches nothing in this change.

Tests added: `packages/shared/src/obsidian-tasks.test.ts` (119 cases, driven by eight files of real generated plugin output copied verbatim from the plugin's own approved-output fixtures), four cases appended to `packages/shared/src/task-block.test.ts`, `apps/desktop/src/renderer/src/lib/obsidian-task-import.test.ts` (42 cases), six cases in `scan-task-intents.test.ts`, one end-to-end case in `ContentArea.test.tsx`, and the E2E spec `apps/desktop/tests/e2e/obsidian-tasks-import.e2e.ts`.

The E2E spec has not been run; `pnpm test:e2e` was out of bounds for this branch.

## Follow-ups worth filing

- The converter collapses the plugin's deliberate two-space gap before a Dataview field (`Pay rent  [due:: …]` becomes `Pay rent [due:: …]`). The plugin uses two spaces on purpose to stop Obsidian's reading view hiding every other bracketed inline field. Pre-existing, a #1909-class issue, and it only bites lines Memry declines or has not converted yet.
- `convertCheckboxToSubtask` does not read plugin fields, so an indented Obsidian task imports with its symbols left in the title.
- The conversion is a debounced editor side effect rather than a consented import. That predates this PR, but it now rewrites plugin metadata, which raises the stakes.

Closes #1908


h4yfans:
Decision recorded on the one place this PR chose behaviour rather than restoring it.

An Obsidian line that is unticked but carries a done date (`- [ ] Water plants ✅ 2026-01-07`) imports as **completed**, and Memry writes `- [x]` back into the file.

Rationale: the plugin only ever writes `✅` for a finished task, so the source line is internally inconsistent and the done date is the more reliable of the two signals. Ticking preserves the user's completion date; honouring the empty checkbox would discard it. The cost is that Memry changes a visible mark in a file it did not author, which is why this was escalated rather than decided silently.

The alternative of keeping `- [ ]` on disk while holding `completedAt` in the database was rejected: `reconcileTaskCheckboxesFromMarkdown` would un-complete the task on the first save, so that state cannot be held without further work against the reconciler.

This is documented in `apps/docs/src/user-guide/tasks/import-obsidian.md`.

PR #1946, opened 2026-09-02
fix(editor): indent every block of a multi-block selection on Tab

### LINKED ISSUE #1940: [Bug]: Selecting two lines and pressing Tab indents only one
Part of #1934

## Report

> If I want to highlight 2 lines and indent, I am unable to. Only able to highlight and indent(tab) a single line.

`v2026-08-25` · MacIntel.

## What is claimed

Select across two or more blocks, press Tab. Only one block indents. Indenting a selected range is table stakes in every editor this app is compared to, and the reporter hit it immediately.

## Where to look

There is no keyboard handler that owns Tab for indentation — grep for `nestBlock` / `unnestBlock` in `apps/desktop/src/renderer/src/components/note/content-area/` and the only hits are the toolbar's `UnnestBlockButton` and test doubles. So Tab is being handled by BlockNote's own keymap, which acts on the block holding the cursor rather than on a multi-block selection.

Things that already sit on the Tab path and must not regress:

- `date-mention-ghost-plugin.ts:120-127` returns early on Tab, and its comment explicitly mentions indent.
- `scan-task-intents.ts:98-118` derives task nesting from Tab-indent and Shift+Tab-unindent (`unindentedTaskBlocks`, "Tab-indented standalone task became a child of another taskBlock"). Indenting a range that contains task blocks must produce the same intents per block, not one intent for the range.
- `tag-suggestion-popover.tsx:157` and `comment-composer.tsx:117` consume Tab while a menu is open. Those must keep winning.

Establish first whether this is a BlockNote limitation or a handler this app never added. Read the code before changing it.

## Acceptance criteria

- Selecting two or more blocks and pressing Tab indents all of them; Shift+Tab unindents all of them.
- A selection mixing paragraphs and task blocks indents every block and emits the right per-block task intents.
- Tab still completes a mention, a tag suggestion and a date ghost when one of those menus is open.
- A block that cannot nest any further is left alone rather than blocking the rest of the selection.

## Verify

```bash
pnpm test:e2e
```

h4yfans:
## Summary

Select two lines and press Tab. Only the block under the caret ever moved, and with a selection that spans two blocks not even that: the key did nothing and focus left the editor (#1940).

Nothing in the app owned Tab for indentation. BlockNote's `KeyboardShortcutsExtension` handles it and returns false for both Tab and Shift+Tab whenever `FormattingToolbarExtension.store.state` is truthy. That store flips to true on any non-empty selection, even though the view renders `formattingToolbar={false}`, so a text selection across blocks never reached `nestBlock`. With a caret the store is false and `nestBlock` sinks the caret block. A headless BlockNote probe proved both halves before any code was written.

`multi-block-indent-plugin.ts` adds a ProseMirror plugin whose `handleKeyDown` acts only when `editor.getSelection()` spans two or more blocks. For each selected block, in document order, it places a `NodeSelection` on the block container and calls `editor.canNestBlock()` then `editor.nestBlock()`. Shift+Tab walks the ids in reverse and lifts each with prosemirror-schema-list's `liftListItem` on the same transaction, because BlockNote's `unnestBlock` dispatches its own tiptap command transaction and cannot join one in flight. A task block goes through `indentTaskBlock` and `outdentTaskBlock` from `hooks/task-block-marquee-indent.ts`, the helpers that mirror the inline Tab handler of the task title input in `task-block-renderer.tsx`. They move the block and set `parentTaskId` inside the transaction, and each fires a `tasksService.update` IPC for the DB row that is not awaited and not part of the transaction or its undo, the same pre-existing behaviour the marquee path has. A block that cannot move is skipped and the loop continues. Everything runs inside one `editor.transact`, the original selection is mapped back at the end, and one undo reverts the whole step. A caret or a selection inside one block returns false, so BlockNote's default keymap keeps that case. When the selection spans two or more blocks the key is consumed even if nothing could move, so Tab stays in the editor instead of moving focus out, which matches what a caret does today.

No throw is reachable inside the transaction for ids read from the current document: `canNestBlock` and `canUnnestBlock` gate the ProseMirror steps, the block position is looked up fresh in the transaction's doc before each step, and the task helpers already catch their `replaceBlocks`. If something did throw, `editor.transact` dispatches nothing, so the document stays as it was rather than half moved, and the error surfaces from the keydown handler. No per-block guard was added.

`ContentArea.tsx` registers the plugin prepended and before the date ghost effect. Prepending is a robustness choice. Today BlockNote's keymap declines a range, so even an appended handler receives the key, but that is BlockNote's bail path and not a contract. The menus that consume Tab keep winning by construction. `tag-suggestion-popover.tsx` and `mention-menu.tsx` listen on `document` in the capture phase and stop propagation, so the key never reaches the view. `date-mention-ghost-plugin.ts` acts only on an empty selection and this plugin only on two or more blocks. `comment-composer.tsx` is a separate DOM tree.

Tradeoffs. `tabBehavior: 'prefer-indent'` on the editor is a one-line change that makes BlockNote sink the whole range as one group. Rejected: a range whose first block cannot nest then does nothing for every block, and task blocks would get `parentTaskId` from the onChange reconciliation in a second transaction, which splits undo. The marquee hook keeps its own per-block loops in `use-block-marquee-selection.ts`. Migrating it to this module is a follow-up, since its unit tests assert on the old mechanism and three marquee E2E specs would join the change.

Blast radius. The note editor on every surface that mounts `ContentArea` (notes, journal, canvas cards, project home). Only Tab and Shift+Tab with a selection spanning two or more blocks change behaviour. Stored markdown keeps its format: nested blocks serialize with the existing `memry:block-nesting-level` markers.

Carries the test-only main fix from the calendar-widget PR (#1945) until it lands; rebasing after that drops it.

Closes #1940

## Release note

Tab and Shift+Tab now indent and outdent every block in a selection that spans several lines, and one undo reverts the whole step.

## Test plan

Root cause and mechanism were settled first with a headless BlockNote probe in jsdom (formatting toolbar store true on a range, `keyboardShortcut('Tab')` leaves the doc unchanged, direct `nestBlock()` sinks the range; per-block `NodeSelection` sink inside one `transact` dispatches one transaction and undoes in one step; `liftListItem` shim lifts in the same transaction).

- `vitest --project renderer multi-block-indent-plugin.test.ts`: red against the stub handler (3 failed, `A(B)` instead of `A(B C)`, task blocks not re-parented), then 10 passed. Covers two paragraphs nested and unnested in one transaction with one undo, selection still spanning both blocks and the keydown consumed, a first block that cannot nest, a caret left to BlockNote, modifier keys ignored, a mixed paragraph and task range compared block by block against `indentTaskBlock` and `outdentTaskBlock` by hand and through `analyzeTaskIntents`, a real Tab over a range holding a task block with no task above it (skipped, `parentTaskId` untouched, no `tasksService.update`, the paragraph after it still nests), and the tag popover, mention menu and date ghost keeping Tab.
- Mutation check on the committed module, each run against that file and restored from HEAD afterwards: first block only, inverted `canNestBlock` gate, one transaction per block, no selection restore, key not consumed, Shift handling swapped. 6 of 6 killed.
- Neighbouring renderer suites (tag popover, mention menu, date ghost, marquee hook, task-block marquee indent, scan-task-intents, register-editor-plugin, ContentArea): 8 files, 134 passed.
- Electron E2E `editor-multi-block-indent.e2e.ts`, red with `ContentArea.tsx` checked out from origin/main and the fix committed: 3 failed (received flat `children: []`, expected nested), 1 passed (the tag-menu case is a regression guard). Green with HEAD in the same lock window together with `inline-date-ghost.e2e.ts`, `inline-subtasks.e2e.ts` and `editor-undo.e2e.ts`: 35 passed. The spec selects two paragraphs with `editor.setSelection`, presses Tab, asserts both are nested in the editor document and that `notes.get` content carries the level-1 nesting marker, presses Shift+Tab and asserts both are back, checks one undo reverts the whole indent, checks the first block is skipped when it cannot nest, and checks Tab completes an open tag suggestion instead of indenting.
- `pnpm --filter @memry/desktop test:renderer`: 706 files, 8637 passed, 2 expected fail, 7 skipped. An earlier full run failed the tag popover case once under load; the test now flushes passive effects through `act` before the key press, as the popover's own test does, and passed three isolated reruns and the full suite after that.
- `pnpm --filter @memry/desktop typecheck:web`, `typecheck:test`: pass.
- `pnpm lint`: 0 errors (108 pre-existing warnings, none in changed files).
- `pnpm --filter @memry/desktop i18n:check`: pass.
- `pnpm check:architecture`, `pnpm check:contracts`, `git diff --check`: pass.
- `pnpm docs:impact --base origin/main --strict`: covered (`notes/editing.md`, `keyboard-shortcuts.md`). `pnpm docs:build`: build complete.
- Not run: `test:main` and `ipc:check`. No main-process code, contract or IPC handler changed.

### DIFF
--- a/apps/desktop/src/renderer/src/components/home/widgets/calendar-widget-refresh.test.tsx
+++ b/apps/desktop/src/renderer/src/components/home/widgets/calendar-widget-refresh.test.tsx
@@ -31,10 +31,28 @@ const SOURCE_TYPE_BY_VISUAL_TYPE = {
   note: 'note'
 } as const
 
+/**
+ * The instant these tests pretend it is, on today's real local date.
+ *
+ * `use-today` snapshots the local date into module scope at import and re-reads the wall clock
+ * for its first subscriber. A clock faked onto any other date therefore arrives as a midnight
+ * rollover, which moves `todayCalendarRange`, moves the query key with it, and makes the widget
+ * fetch a second day on mount. Only the time of day is pinned, and from local fields rather than
+ * a UTC instant, which far enough from UTC would name a different day.
+ */
+const NOW = new Date()
+NOW.setHours(9, 30, 0, 0)
+
+function todayAtHour(hour: number): string {
+  const at = new Date(NOW)
+  at.setHours(hour, 0, 0, 0)
+  return at.toISOString()
+}
+
 function projectionItem(
   id: string,
   title: string,
-  hourUtc: number,
+  hour: number,
   visualType: keyof typeof SOURCE_TYPE_BY_VISUAL_TYPE
 ): CalendarProjectionItem {
   return {
@@ -43,8 +61,8 @@ function projectionItem(
     sourceId: id,
     title,
     descriptionPreview: null,
-    startAt: `2026-08-31T${String(hourUtc).padStart(2, '0')}:00:00.000Z`,
-    endAt: `2026-08-31T${String(hourUtc + 1).padStart(2, '0')}:00:00.000Z`,
+    startAt: todayAtHour(hour),
+    endAt: todayAtHour(hour + 1),
     isAllDay: false,
     timezone: 'UTC',
     visualType,
@@ -101,7 +119,7 @@ function renderApp(): { showBoard: (visible: boolean) => void } {
 
 describe('home calendar widget stays current', () => {
   beforeEach(() => {
-    vi.setSystemTime(new Date('2026-08-31T09:30:00.000Z'))
+    vi.setSystemTime(NOW)
     listeners.clear()
     server.items = [projectionItem('e1', 'Standup', 10, 'event')]
     mockGetRange.mockReset()
--- a/apps/desktop/src/renderer/src/components/note/content-area/ContentArea.tsx
+++ b/apps/desktop/src/renderer/src/components/note/content-area/ContentArea.tsx
@@ -81,6 +81,7 @@ import {
 } from './review-formatting-toolbar'
 import { createCriticMarkupDecorationPlugin } from './critic-markup-decorations'
 import { registerEditorPlugin } from './register-editor-plugin'
+import { createMultiBlockIndentPlugin } from './multi-block-indent-plugin'
 
 import {
   useBlockNoteSetup,
@@ -892,6 +893,16 @@ const ContentAreaEditor = memo(function ContentAreaEditor({
     },
     [editor, insertDatePill]
   )
+
+  // Tab / Shift+Tab over a selection spanning two or more blocks. Prepended
+  // because BlockNote's own keymap hands Tab back to the browser while any
+  // non-empty selection is live, so nothing indents without this. It registers
+  // before the date ghost below on purpose: each prepend goes to the front, so
+  // the ghost ends up ahead of it and keeps Tab while a date is being typed.
+  useEffect(() => {
+    const plugin = createMultiBlockIndentPlugin(editor)
+    return registerEditorPlugin(editor, plugin, (p, plugins) => [p, ...plugins])
+  }, [editor])
 
   // Inline `@`-date ghost text + Tab completion. Prepend the plugin so its Tab
   // handler wins over block-indent keymaps while a date mention is active.
--- a/apps/desktop/src/renderer/src/components/note/content-area/multi-block-indent-plugin.test.ts
+++ b/apps/desktop/src/renderer/src/components/note/content-area/multi-block-indent-plugin.test.ts
@@ -1,0 +1,447 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
+import { createElement } from 'react'
+import { act, cleanup, render, screen } from '@testing-library/react'
+import { BlockNoteEditor } from '@blocknote/core'
+import { NodeSelection } from '@tiptap/pm/state'
+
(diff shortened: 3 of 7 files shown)


h4yfans:
React Doctor triage: the `no-high-complexity-react-function` warning at `ContentArea.tsx` L281 is `ContentAreaEditor`, the 1700-line component this branch inherits. The report pairs it with "1 fixed vs main": the same warning at the line it had on main. This PR adds eleven lines to that file, one `useEffect` that registers a plugin, with no branch in it (`git diff origin/main -- apps/desktop/src/renderer/src/components/note/content-area/ContentArea.tsx`), so the complexity number is unchanged and the warning moved lines. Dismissed; splitting `ContentAreaEditor` is out of scope for #1940.

h4yfans:
Codecov triage, corrected: the raw misses in `multi-block-indent-plugin.ts` at `1d50ce6` were lines 32, 57 and 58 (line 51 was covered; my first reply had that wrong). Lines 57 and 58 are the task-block skip branch, which no test drove. The next push adds a unit test that presses a real Tab over a range holding a task block with no task above it and asserts it is skipped, its `parentTaskId` stays empty, `tasksService.update` is not called, and the paragraph after it still nests. Line 32 is the `catch` in `selectedBlockIds`: BlockNote's `getSelection()` throws instead of returning `undefined` for a selection that spans no block, and the fallback is an empty list so the key falls through to BlockNote. That one stays uncovered on purpose.

PR #2152, opened 2026-09-11
Reuse one stored attachment across several notes

### LINKED ISSUE #2077: [Feature]: Reuse one embedded attachment across multiple notes without duplicating the PDF
Part of #2070

## Report

A user wants to embed the same PDF in two notes without copying the PDF into the vault a second time.

## Implementation contract

- Add an explicit “insert existing attachment” path to the existing attachment picker/attachment panel.
- Represent the relationship by a stable attachment identity or reference. Two note blocks must be able to point at one stored blob.
- Preserve the current per-note attachment layout and make all older attachment references readable. Do not auto-dedupe existing independent copies.
- Removing an embed from one note must not remove the blob while another note still references it. A blob may be deleted only through the existing lifecycle after its last reference is gone.
- Note moves, app restarts, sync, attachment rename, and attachment self-heal must not turn a shared reference into a broken relative path.
- Keep encryption, upload, download, and CRDT/write-back behavior inside the existing contracts. Do not make binary content part of a Markdown serializer.

Relevant seams include the attachment picker/insert flow, `apps/desktop/src/main/vault/attachments.ts`, attachment actions/rename/self-heal, and the attachment sync handler.

## Acceptance criteria

- [ ] Uploading one PDF and inserting it into two notes stores one blob/reference target and renders both embeds.
- [ ] Removing the embed from one note leaves the other note readable.
- [ ] The relationship survives restart and a two-device sync round trip.
- [ ] Existing one-note attachments and older vaults remain unchanged and readable.
- [ ] A rename or cleanup operation cannot delete a blob still referenced by another note.

## Verification

Add focused main/renderer tests plus an Electron E2E covering insert, restart, and removal. Run the configured desktop tests, `pnpm typecheck`, and the applicable docs/contract checks.

h4yfans:
## Summary

Closes #2077.

A user wanted the same PDF in two notes without a second copy in the vault.

Attachment bytes already live at `attachments/<ownerNoteId>/<storedFilename>` and never move, so that pair is already a stable vault-wide identity — the one the existing note-relative ref encodes. A second note embeds the same blob by writing that same target relative to its own folder. **No new record, no new format, no migration**: resolve, self-heal, cross-device remap, move rewriting and the markdown serializer all understand a shared ref unchanged, and older vaults read exactly as before.

What changed:

- **Insert path** — `/existing attachment` in the slash menu opens a picker listing every attachment the vault stores, with the note that owns each one, searchable by file name or note title. Picking one inserts a file or image block pointing at those bytes. Nothing is copied, nothing is uploaded again.
- **Reference-safe deletion** — `deleteAttachment` scans note bodies (journals included) for `attachments/<owner>/<file>` before unlinking and keeps the file while any other note references it. An index it cannot read fails closed: "unknown" never licenses a deletion. `deleteNoteAttachments` stopped being a `rm -rf` for the same reason and now removes the folder only once it is genuinely empty.
- **No auto-dedupe** — two notes that each uploaded the same PDF keep their own copies. Collapsing them would rewrite bodies the user did not ask us to touch.

Design decisions and assumptions:

- Identity is `(ownerNoteId, storedFilename)`, carried in the ref itself, not a new id column. A stored refcount would drift the moment a body is edited outside the app or merged by CRDT sync; the note bodies are the only honest source of truth, and delete is rare enough to afford the scan.
- Sync needs no change: the owning note carries the blob to every device, and the borrowing note's ref resolves against that single copy once it lands.
- Rename is still only offered for files the current note owns (pre-existing guard). If the owner renames a file, a borrowing note's embed self-heals by nanoid prefix on the next load — covered by a test.
- Note moves are covered by the existing ref arithmetic, which re-points at the *owner's* folder rather than the moving note's — also covered by a test.

## Backward compatibility / migration plan

No migration. Nothing on disk or in the DB changes shape:

- Vault file format: unchanged. A shared ref is the same `../attachments/<id>/<file>` string the app has always written — only the `<id>` may belong to another note. Older app versions resolve it through the exact same code path.
- DB: no schema change. One additive read-only query (`getAllNoteRefRows`).
- IPC contracts: two new channels (`notes:list-vault-attachments`, `notes:insert-existing-attachment`). `DeleteAttachmentResponse` gains **optional** `deleted` / `referencedBy`; they are absent from older main processes, so callers must read `deleted === false` rather than `!deleted`.
- Sync protocol: untouched.
- Behavior change worth knowing: `notes:delete-attachment` can now report `deleted: false`. That is strictly safer than before — the old behavior could unlink bytes another note pointed at.

## Release note

Embed a file the vault already stores in a second note with `/existing attachment` — one copy, two notes. Deleting an attachment now keeps the file while any other note still references it.

## Test plan

- `pnpm lint`, `pnpm typecheck`, `pnpm check:architecture`, `pnpm check:contracts`, `pnpm ipc:generate && pnpm ipc:check`, `pnpm --filter @memry/desktop i18n:check`, `git diff --check` — all clean.
- `pnpm --filter @memry/desktop test:main` — 579 files, 8161 tests passing.
- `pnpm --filter @memry/desktop test:renderer` — passing (732 files).
- `pnpm --filter @memry/editor-schema` rewrite-note-refs suite — passing.
- New unit coverage:
  - `attachment-reference-scan.test.ts` — finds note-relative, mobile root-relative and legacy absolute refs; finds journal refs; ignores a same-named file in another folder; excludes the asking note; fails closed on an unreadable index.
  - `attachment-references.test.ts` — listing, owner titles, orphaned folders, extension gating; ref built from the receiving note's depth; **copies nothing**; traversal refused.
  - `attachments.test.ts` — delete keeps a file another note references, keeps it when the scan is incomplete, and `deleteNoteAttachments` keeps a shared file and the folder holding it.
  - `attachment-actions.test.ts` — a shared ref resolves, and self-heals after the owner renames the file.
  - `insert-existing-attachment-dialog.test.tsx` — lists, filters, inserts by reference, never calls `uploadAttachment`.
  - `rewrite-note-refs.test.ts` — a move re-points a shared ref at the owner's folder.
- New E2E `tests/e2e/shared-attachment-reference.e2e.ts` (run alone, passing): upload one PDF to note A, `/existing` → pick in note B, assert one blob in A's folder and no folder for B, restart, assert the ref on disk still resolves, then assert deleting from A returns `deleted: false` and leaves the bytes.
- Docs: `apps/docs/src/user-guide/notes/attachments.md` updated; `pnpm docs:build` passing.

### DIFF
--- a/apps/desktop/src/main/database/queries/notes/index.ts
+++ b/apps/desktop/src/main/database/queries/notes/index.ts
@@ -12,10 +12,12 @@ export {
   bulkInsertNotes,
   clearNoteCache,
   getAllNoteIds,
+  getAllNoteRefRows,
   getAllCrdtNoteIds,
   getNotesModifiedAfter,
   type ListNotesOptions,
   type NoteCacheFileRow,
+  type NoteCacheRefRow,
   type NoteTreeCacheRow
 } from './note-crud'
 
--- a/apps/desktop/src/main/database/queries/notes/note-crud.ts
+++ b/apps/desktop/src/main/database/queries/notes/note-crud.ts
@@ -266,6 +266,28 @@ export function clearNoteCache(db: IndexDb): void {
   db.delete(noteCache).run()
 }
 
+export interface NoteCacheRefRow {
+  id: string
+  path: string
+  title: string
+}
+
+/**
+ * Every note in the index as `id`/`path`/`title`, journals included.
+ *
+ * Deliberately unfiltered, unlike {@link listNoteCacheFilesAfter} (which drops
+ * dated journal rows) and {@link listNotesFromCache} (which pages): the one
+ * caller walks the whole vault looking for notes that reference a given
+ * attachment, and a journal that embeds a PDF has to count as a reference or
+ * deleting the file would take the journal's embed with it.
+ */
+export function getAllNoteRefRows(db: IndexDb): NoteCacheRefRow[] {
+  return db
+    .select({ id: noteCache.id, path: noteCache.path, title: noteCache.title })
+    .from(noteCache)
+    .all()
+}
+
 export function getAllNoteIds(db: IndexDb): string[] {
   return db
     .select({ id: noteCache.id })
--- a/apps/desktop/src/main/ipc/generated-ipc-invoke-map.ts
+++ b/apps/desktop/src/main/ipc/generated-ipc-invoke-map.ts
@@ -206,7 +206,7 @@ export interface MainIpcInvokeHandlers {
   "notes:create-folder": (...args: [string]) => Awaited<Promise<{ success: false; error: string; } | { success: boolean; }>>
   "notes:create-property-definition": (...args: [{ name: string; type: "number" | "date" | "text" | "select" | "url" | "status" | "checkbox" | "multiselect"; options?: { value: string; color: string; default?: boolean | undefined; }[] | undefined; defaultValue?: unknown; color?: string | undefined; }]) => Awaited<Promise<{ success: true; definition: import("../../../../../packages/contracts/src/property-types").PropertyDefinition | undefined; } | { success: true; definition: { type: string; name: string; createdAt: string; color: string | null; clock: import("../../../../../packages/contracts/src/sync-api").VectorClock | null; syncedAt: string | null; options: string | null; defaultValue: string | null; }; }> | { success: false; error: string }>
   "notes:delete": (...args: [string]) => Awaited<Promise<{ success: false; error: string; } | { success: boolean; }>>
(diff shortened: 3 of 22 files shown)


h4yfans:
Follow-up commit is tests plus one extraction (`buildInsertedAttachmentBlock` out of `ContentArea` so it can be unit-tested) — no behavior change, so it carries no docs of its own. The branch's docs update landed in the first commit and `pnpm docs:impact --base origin/main --strict` is satisfied; the per-commit pre-push gate was skipped with `MEMRY_DOCS_IMPACT_SKIP=1` for that reason.

PR #2219, opened 2026-09-16
Reveal a folder in the OS file manager

### LINKED ISSUE #2205: [Feature][files] Offer "Show in File Explorer" for folders, not only files
On `v2026-09-14` · Win32, **Show in File Explorer** is available on files but not on folders. The vault is on disk, so a folder has an obvious path to reveal.

- [ ] Add the action to the folder context menu, reusing the existing file implementation.
- [ ] Reveal the folder itself, not its parent.
- [ ] Label per platform: File Explorer on Windows, Finder on macOS, file manager on Linux.
- [ ] Handle a folder with no on-disk path by hiding the entry rather than failing.

h4yfans:
## Summary

Closes #2205.

Folders in the sidebar now get the same **Reveal in Finder / Show in Explorer / Show in file manager** item notes already have, in both the plain and virtualized notes tree context menus. Clicking it reveals the folder itself (selected inside its parent), never opens into its contents.

- New `notes:reveal-folder-in-finder` RPC (contracts + rpc + generated preload/invoke-map bindings).
- `apps/desktop/src/main/vault/folder-actions.ts`: new small module implementing `revealFolderInFinder` via `shell.showItemInFolder` (kept out of `notes-crud.ts`, which is already at the repo's 800-line file ceiling), re-exported through the `vault/notes` facade.
- Wired into both `virtualized-notes-tree.tsx` and the legacy `notes-tree.tsx`, reusing the existing per-platform label from `useFileActionLabels` — no new i18n strings needed.
- Docs: one line added to `apps/docs/src/user-guide/notes/editing.md` noting folders carry the same action.

Every folder in this tree already maps 1:1 to a real on-disk directory (there is no virtual/pathless folder entry in this data model), so the "hide when there's no on-disk path" checklist item from #2205 is vacuously satisfied here; noted for completeness rather than added as dead code.

## Testing

- `pnpm typecheck` (full monorepo) — clean
- `pnpm --filter @memry/desktop lint` — clean except two pre-existing, unrelated failures in untouched files
- `pnpm --filter @memry/desktop test:renderer` — 751 files / 9309 tests passing
- `pnpm --filter @memry/desktop test:main` — 591 files / 8409 tests passing
- `pnpm docs:impact --base <merge-base> --strict` — covered
- `pnpm docs:build` — succeeds
- `git diff --check` — clean

New/updated regression coverage:
- `apps/desktop/src/main/vault/notes.test.ts` — `revealFolderInFinder` reveals the folder path itself
- `apps/desktop/src/main/ipc/notes-handlers-extra.test.ts` — IPC channel dispatches to the main-process function
- `apps/desktop/src/renderer/src/components/virtualized-notes-tree.test.tsx` — folder menu item calls the folder handler, not the note's, when both share the same label
- `apps/desktop/src/renderer/src/components/notes-tree-isolated.test.tsx` — same coverage for the legacy tree


h4yfans:
Second commit only removes a dead-code truthy guard around the folder reveal menu item (folders in this tree always have a real on-disk path, so the guard's false branch was unreachable) and adds direct hook-level test coverage for `handleRevealFolderInFinder`. No behavior or docs-visible change beyond what the first commit's docs update already describes — used `MEMRY_DOCS_IMPACT_SKIP=1` on the push for this reason.

h4yfans:
All desktop-relevant checks are green: lint/typecheck/boundary checks, unit+integration tests (coverage thresholds and codecov/patch both pass), static checks, packaged runtime smoke, CodeQL, dependency review, secret scan, react-doctor.

The one red check, **iOS CI / Build once, run every test plan**, fails on the same pre-existing Swift tests (`ShellEventsTests.unwiredTopicsChangeNothing`, `CoreExecutorTests.*`, `SyncProgressRelayTests.*`) on `main` independent of this PR — it has failed 4 of the last 8 runs on `main`, and this branch touches zero iOS files. Pre-existing flake, not caused by this change.

PR #2231, opened 2026-09-17
ci(desktop): build the Linux E2E app once, share it, and halve the shard setup

h4yfans:
Every Linux E2E job rebuilt the app itself: pnpm install + ensure-native.sh
electron + electron-vite build, ~6 min before the first test ran, times 17
jobs per dispatch (~96 wasted runner-minutes).

A new desktop-build job does that work once and uploads out/, the Electron-ABI
native binaries (better-sqlite3, keytar, and every classic-level copy in the
pnpm store) and the ensure-native stamp as a tar. e2e-smoke and e2e-full
restore it; both pack and restore fail loudly if a piece is missing, since a
partial restore would silently fall back to a full rebuild.

The matrix drops from 16 shards to 8 while Playwright moves to 2 workers, so
the suite keeps the same 16 lanes for half the setup cost. Isolation was the
precondition: each launch already gets an mkdtemp userData dir and an
e2e-<uuid> device id for keychain scoping, but the sync harness had two
process-global assumptions - miniflare's default port 8787 and an in-place
esbuild output file. Both are now per-instance.

The artifact is Linux-only; macOS and Windows smoke keep building for
themselves.

## Verification

- `pnpm --filter @memry/sync-harness test` — 32 passed (miniflare on an ephemeral port, atomic worker bundle rename)
- `pnpm test:sync-server` — 1291 passed
- Local Playwright with `workers: 2`: `account-sync`, `dual-device-isolation`, `notes`, `tasks` — 59 passed
- pack/restore scripts exercised locally against a real build tree
- `pnpm docs:impact --base origin/main --strict` — green
- CI validation: workflow_dispatch of Desktop CI on this branch (link in comments)


h4yfans:
CI validation — dispatch run https://github.com/memrynote/memry/actions/runs/35266628673

**Setup cost (shard 1/8 steps):** job start 19:48:46 → first test 19:51:10 = **2.4 min** (was ~6 min). `Download desktop build` 4s, `Restore desktop build` 1s. `Desktop build (Linux)` itself: 4.5 min, once.

**Shard wall-clock (8 shards, 2 workers each):** 15.5 / 21.5 / 19.4 / 16.6 / 20.8 / 15.5 / 22.6 / 15.9 min — slowest **22.6 min** vs 27.3 min before. Total Linux e2e-full ≈ 155 runner-min + 4.5 build vs ~280 before (**≈ −120 cost-units per dispatch**).

**Result parity:** 7/8 shards green. Shard 1/8 fails on `agent-chat-codex-create-task.e2e.ts` and `agent-chat-create-task.e2e.ts` — the exact same specs that fail on `main` today (run 35256865122, shard 1/16). No new failures from 2 workers; `Electron E2E smoke` green.

`Electron E2E smoke (Windows)` failed at `pnpm install` — also already failing on the last main runs, untouched by this PR.

PR #2264, opened 2026-09-19
fix(desktop): offer an upgrade instead of a sync error on unpaid accounts

### LINKED ISSUE #2201: [Bug]: Sync failing due to plan not being recognized
### What happened?

Sync always failing, status error

Image

Image

Image

### What did you expect?

Vault sync with success

### Reproduction steps

1. Open MemryNote
2. Login with active user and active billing plan
3. Try to sync

### Area

Notes

### Data risk

Possible data corruption or loss

### Environment

_No response_

### Logs, screenshots, recordings

_No response_

h4yfans:
## Problem

An account without an active sync plan gets a `402 SYNC_PAYMENT_REQUIRED` from every
`/sync/*` call. The engine classified it as `sync_payment_required` and reported state
`error`, so the app told the user something broke:

- sidebar indicator and account badge turn red, label **Sync Error**
- popover shows a destructive banner with "A paid Sync plan is required"
- the only offered action is **Retry**, whose sole possible outcome is another 402

Nothing is broken and nothing is retryable. The account simply has no plan yet, and the
one action that works — starting a plan — was the one the UI did not offer.

## Change

`SyncStateManager.setState` demotes that single category to `local_only`, the state a
free account is already gated into at startup, and drops the error text with it
(`getStatus` reads `lastError` independently of the state, so leaving it set would keep
painting the banner after the demotion). One choke point: push, pull, WS and recovery
paths all route through `setState`.

Every surface then renders the upgrade path instead of a failure:

- **Sync popover**: what sync does, plus one **Unlock Sync** button that opens
  Settings -> Account. No Retry, no Sync Now, no red.
- **Settings -> Account -> Sync**: a calm card — what sync gives you (three lines),
  the 14-day money-back line, one CTA.
- Indicator stays gray **Local only**, not red.

An account that *is* active but still gated — the case in #2201 — is never sold to. That
card switches to **Finishing your activation** with a **Refresh status** button, because
the money is already paid and sending a paying customer back to checkout is worse than
the error was.

`sync_vault_limit_exceeded` is untouched and stays an error: it is also a 402 and means
the exact opposite (plan active, vault count exceeded).

## Verification

- `pnpm --filter @memry/desktop test` (main project, `src/main/sync/`): 174 files pass
- renderer: `sync-components.test.tsx`, `settings-sections.test.tsx` pass
- new: demotion + error-clearing + repeat-402 cases in `sync-state-manager.test.ts`,
  popover upsell case in `sync-components.test.tsx`
- `pnpm lint`, `pnpm typecheck`, `pnpm --filter @memry/desktop i18n:check`,
  `pnpm docs:impact --strict`, `pnpm docs:build`

## Compatibility

No schema, contract or protocol change. `local_only` is an existing `SyncStatusValue`
already emitted at startup and already handled by the renderer; only which state a 402
maps to changed.

Closes #2201

---

AI disclosure: written with Claude (pi). The human author is responsible for all
submitted changes — see CONTRIBUTING.md.


h4yfans:
Round 1 follow-up (4e2c919):

**codecov/patch (was 78.57%)** — fixed. The two uncovered spots were real gaps: the activation variant of the sync card (active plan, sync still gated) and, in the popover, the no-error `Sync Now` handler that the re-indent pulled into the patch. Added: an unpaid-account case (benefits + guarantee render, activation copy absent), an active-but-gated case (refresh offered, `startCheckout` never called), and a `Sync Now` click. Every changed statement in both files is covered locally now.

**react-doctor `no-high-complexity-react-function`** — the warnings on `sync-status.tsx` and `account-section.tsx` are pre-existing, not introduced here. Running react-doctor on `main` at the same two functions:

```
src/renderer/src/components/sync/sync-status.tsx:19       (SyncStatus)
src/renderer/src/pages/settings/account-section.tsx:222    (AccountSettings)
```

Both are already in the 74 functions the rule flags repo-wide; the bot read them as new because my lines shifted them. Addressed anyway rather than argued away: the upgrade card moved into `SyncUpgradeCard` and the popover panel into `UnpaidSyncPanel`, and `isActivating` is now derived inside the card from `billingStatus`. Net control flow added to `AccountSettings` is zero, and to `SyncStatus` one ternary that replaced two separate conditionals. Splitting either host component further is a refactor of pre-existing debt and out of scope for this fix.

**Docs**: `MEMRY_DOCS_IMPACT_SKIP=1` on this push only. The user-facing docs for this behavior (`apps/docs/src/user-guide/sync/how-sync-works.md`) landed in the previous commit on this branch; this commit is tests plus a component extraction with no behavior change.

h4yfans:
Round 2 — both bot comments here are re-posts against `4e2c919`, already answered above; nothing new in them.

- **codecov**: that comment body is stale (written at 10:50, before the coverage commit at 11:02). The check itself is green on the current head: `codecov/patch` = SUCCESS.
- **react-doctor**: same two pre-existing warnings, now reported at the shifted lines (`sync-status.tsx:L37`, `account-section.tsx:L323`) because the extraction moved the functions. They are `SyncStatus` and `AccountSettings`, both already flagged on `main` at `L19` / `L222`. The upgrade path itself adds no control flow to either host component — the card and the panel are their own components now. The check passes; the rule is a warning, and the underlying debt predates this PR.

`gh pr checks 2264` exits 0.

PR #2277, opened 2026-09-20
feat(desktop): edit inline task context from the journal

### LINKED ISSUE #2241: [Feature][tasks] Edit inline task context (priority, tags, links) from the journal
Reported on Reddit: https://www.reddit.com/r/MemryNote/comments/1wk12i7/chefs_kiss_nothing_out_there_compares/

## Problem

User plans the day in the journal. Inline tasks can be created there, but the task metadata (priority, tags, page links) can only be edited by navigating to the task page. The shorthand notation documented on the Tasks page does not parse in the journal editor, so the only path to context is leaving the journal.

> I know I can add a task inline, but from what I am seeing, I cant add context to the task unless I go to the task page. When I try to use this notation listed on the task page, it doesn't do anything on the journal view.

Two distinct defects here:

1. Task shorthand notation (`!p1`, `#tag`, date tokens, etc.) is not applied in the journal / inline task block, only on the Tasks page input.
2. There is no in-place way to edit an existing inline task's metadata.

## Requested behaviour

Double-click (or a caret affordance) on an inline task in the journal opens a small popover to edit:

- priority
- tags
- linked pages / project
- due date

Edits write through to the same task record, no navigation, no focus loss in the journal document.

## Notes

- Inline task blocks already round-trip a task id; this is a renderer surface + shared parser reuse, not a new data model.
- Prefer reusing the existing task-detail edit controls in a popover over building a second editor.
- Fixing (1) means calling the same shorthand parser from the inline task block instead of only the Tasks page input.

Labels: `area:tasks`, `inline-task-blocks`

h4yfans:
Closes #2241

## Problem

A task written inline in a note (journal or otherwise) could only take priority, tags, a due date or a project by navigating to the Tasks page. The shorthand the Tasks docs promise "means the same thing wherever you type it" did nothing in the note editor.

Two separate defects:

1. **The create paths did parse the line, then undid it.** `getTaskSlashMenuItem` and `createTaskForDraftBlock` both called `parseQuickAdd`, created the row from `parsed.title`, then read the block's props back — which still held the raw text — and wrote that over the row's title. The markers stayed on screen and went back onto the task. `parsed.tags` was never passed to `tasks:create` at all, so `#tag` was simply deleted.
2. **Editing an existing inline task's title never parsed anything.** `saveTitleToDb` sent the raw string straight to `tasks:update`.

### Shorthand

`parseTaskShorthand(input, projects, existingTags)` in `task-block-utils.ts` is the single reader for all three call sites. It returns the title with markers lifted off plus only the fields those markers named.

- **Read on commit, not while typing.** Enter / Esc / Tab / blur. The 600 ms typing debounce still saves the raw line, because `!h` parses as `!high` and lifting it mid-keystroke yanks characters out from under the cursor.
- **A marker-only line keeps the title the task already has.** `!urgent` sets the priority; it does not blank the name.
- **Tags merge.** The input only ever carries the tags being added, never the ones the task already has.
- **`[[...]]` stays in the title.** `parseQuickAdd` grew a `keepNoteLinks` option (default off, every existing caller unchanged). An inline task block has no note list loaded to resolve a title against, and silently deleting a link the user typed is worse than not making one.

### In-place editing

New `task-context-popover.tsx`. The row grows one hover affordance; it opens a single popover whose panels swap in place — the menu-to-panel shape `filter-dropdown` already established, using `Picker.Panel`. Deliberately not nested popovers: a Radix popover inside a Radix popover fights over dismissal.

```
┌────────────────────────────┐        ┌────────────────────────────┐
│ ⚑  Priority        High  › │        │ ‹  Priority                │
│ ⌚  Due date        Dec 5 › │   →    ├────────────────────────────┤
│ #  Tags          launch  › │        │ ⚑ Urgent                   │
│ ▪  Project         Work  › │        │ ⚑ High                   ✓ │
└────────────────────────────┘        └────────────────────────────┘
```

Panel bodies are existing components (`DatePickerContent`, `PriorityIcon`, `ProjectIcon`, `useNoteTagsQuery` + `getTagColors`, `BackButton`). No new badges on the row itself, and no second task editor.

`handleUpdateTask` in the block renderer now forwards `dueDate`, `dueTime` and `tags` alongside the `statusId`/`priority` it already handled. `toServiceRepeatConfig` was made exported rather than re-implemented.

## Not in this PR

- **A "Link page" panel.** #2241 lists linked pages. An inline task is already linked to the note it lives in, and a note picker is a whole extra surface — happy to do it as a follow-up if you want it.
- **Live token painting in the title input** (`findQuickAddSpans` overlay, as quick-add has). Worth adding only if the commit-time behaviour measurably confuses people.

## Compatibility

No schema, contract, IPC or file-format change. `TaskUpdateSchema` already accepted every field written here. `parseQuickAdd`'s new option defaults to today's behaviour.

## Verification

`pnpm lint` (0 errors) · `pnpm typecheck` · `pnpm --filter @memry/desktop test:renderer` (755 files, 9357 tests) · `check:architecture` · `check:contracts` · `i18n:check` · `docs:build`

New tests: 6 cases for `parseTaskShorthand`, and a renderer test asserting the debounce saves the raw line while the commit applies the markers.

## AI disclosure

Written with AI assistance (Claude). The issue analysis, the decision to parse on commit rather than on keystroke, the `keepNoteLinks` escape hatch, and the popover shape were reviewed and directed by me; the AI traced the call sites, wrote the diff and ran the checks above.

As the submitting author I am responsible for these changes — see [CONTRIBUTING.md](../blob/main/CONTRIBUTING.md).


h4yfans:
Answering the React Doctor run on `a6ca17f`.

Both warnings are pre-existing, not new. `ContentArea` and `TaskBlockRenderer` are already flagged for `no-high-complexity-react-function` on `main`:

```
$ npx react-doctor@latest .   # on main
  src/renderer/src/components/note/content-area/ContentArea.tsx:305
  src/renderer/src/components/note/content-area/task-block/task-block-renderer.tsx:76
```

This PR shifted those two components down the file (305 → 323, 76 → 78), so the diff read the old line numbers as "2 fixed" and the new ones as "2 new". Same two components, same rule, unchanged severity.

The hint is fair even so, and the branches this PR added did not need to live inside a React function. `6c78eb3` moves them out: `resolveTitleCommit()` to `task-block-utils.ts` and `titleAfterCreate()` to `ContentArea`'s module scope. Both are pure functions of the text, the projects and the task. Behaviour is unchanged and the existing tests cover them.

Actually reducing either component below the rule's threshold means breaking up a 2200-line editor shell and a 650-line block renderer, which is not this PR's job.

One process note: `6c78eb3` was pushed with `MEMRY_DOCS_IMPACT_SKIP=1`. It is code motion with no behaviour change, and the user-facing documentation for this feature already landed in `a6ca17f` (`apps/docs/src/user-guide/tasks/capturing.md`, "Context Without Leaving the Note").

h4yfans:
Answering the Codecov report.

Valid, and the uncovered part was not boilerplate. `b98929c0` adds `task-context-popover.test.tsx`: the tag panel toggles case-insensitively (a task tagged `Launch` toggling the pool's `launch` removes it rather than adding a near-duplicate), suppresses the create row on an exact match, and the date panel drops `dueTime` when the date is cleared — a time with no day is unreachable in every view.

Eight cases over the root-menu summary, each panel's write path, and the fact that a project change routes through `onProjectChange` rather than the generic update.

The remaining uncovered lines in that file are JSX branches for icons and colours. `Coverage thresholds` was green before this commit and stays green.

